PR #50, opened 2014-10-20
Always use workers under the hood to execute commands.

MarcCote:
By default, the number of workers is set to the number of commands to be executed.

Using `-p`, one can set the number of workers to use.


MarcCote:
Ok, now that issue #59 is closed, is something missing for this PR?

MarcCote:
I guess I'll create a new PR. Anyway, I should rebase master first.

Issue #73, opened 2015-01-22
smart_dispatch.py -f doesn't work when prepended with a path


samuelstjean:
Launching a file with commands does not work when the full path is specified.

stjeans1 ~ $ ls commands.txt 
-rw-rw-r-- 1 stjeans1 descotea 9 Jan 22 09:16 commands.txt

# This works 
stjeans1 ~ $ smart_dispatch.py -f commands.txt -q qwork@mp2 launch

# This doesn't
stjeans1 ~ $ smart_dispatch.py -f /home/stjeans1/commands.txt -q qwork@mp2 launch
Traceback (most recent call last):
  File "/home/stjeans1/.local/bin/smart_dispatch.py", line 180, in <module>
    main()
  File "/home/stjeans1/.local/bin/smart_dispatch.py", line 47, in main
    path_job, path_job_logs, path_job_commands = create_job_folders(jobname)
  File "/home/stjeans1/.local/bin/smart_dispatch.py", line 171, in create_job_folders
    os.makedirs(path_job_commands)
  File "/opt/python64/2.7.3/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 20] Not a directory: '/home/stjeans1/commands.txt/commands'

Removing the txt doesn't work, it seems to try to create a folder up to the path, with the script name, which fails since you can't have another folder named after the script.

MarcCote:
@samuelstjean you can use PR #74 until @mgermain merges it.

samuelstjean:
Thanks, I just cd'ed to the directory instead to launch the stuff yesterday instead as a workaround.

PR #74, opened 2015-01-22
Take the basename instead of the path when commandsFile is specified.

MarcCote:
Closes issue #73


mgermain:
We should add unittests for the #73 issue before merging.

PR #80, opened 2015-03-03
Failed command management

### LINKED ISSUE #72: Change the way we handle finished jobs.
When a worker see that a job is finished but that it returned something different than 0 as exit code (ie: the job has crashed) move it in failed_commands.txt instead of finished_commands.txt.

mgermain:
Close #72

### DIFF
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -60,6 +60,27 @@ def main():
         if args.mode == "launch":
             command_manager.set_commands_to_run(commands)
         else:
+            # Verifying if there is failed commands
+            failed_commands = command_manager.get_failed_commands()
+            if len(failed_commands) > 0:
+                FAILED_COMMAND_MESSAGE = """{line}
+                {nb_failed} command(s) are in a failed state. They won't be resumed.
+                Failed commands:
+                {failed_commands}
+                The actual errors can be found in the log folder under:
+                {failed_commands_err_file}
+                {line}
+                """
+                print FAILED_COMMAND_MESSAGE.format(
+                    line="-" * 84,
+                    nb_failed=len(failed_commands),
+                    failed_commands=''.join(failed_commands),
+                    failed_commands_err_file='\n'.join([utils.generate_uid_from_string(c[:-1])+'.err' for c in failed_commands])
+                )
+
+                if not utils.yes_no_prompt("Do you want to continue?", 'n'):
+                    exit()
+
             command_manager.reset_running_commands()
             nb_commands = command_manager.get_nb_commands_to_run()
 
@@ -94,7 +115,7 @@ def main():
         jobs_id = []
         for pbs_filename in pbs_filenames:
             qsub_output = check_output('{launcher} {pbs_filename}'.format(launcher=LAUNCHER if args.launcher is None else args.launcher, pbs_filename=pbs_filename), shell=True)
-            jobs_id += [qsub_output.rstrip()]
+            jobs_id += [qsub_output.strip()]
 
         with utils.open_with_lock(os.path.join(path_job, "jobs_id.txt"), 'a') as jobs_id_file:
             jobs_id_file.writeline(t.strftime("## %Y-%m-%d %H:%M:%S ##\n"))
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -69,8 +69,7 @@ def main():
                 Failed commands:
                 {failed_commands}
                 The actual errors can be found in the log folder under:
-                {failed_commands_err_file}\
-                """)
+                {failed_commands_err_file}""")
                 utils.print_boxed(FAILED_COMMAND_MESSAGE.format(
                     nb_failed=len(failed_commands),
                     failed_commands=''.join(failed_commands),
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -6,6 +6,7 @@ import argparse
 import time as t
 import numpy as np
 from subprocess import check_output
+from textwrap import dedent
 
 from smartdispatch.command_manager import CommandManager
 
@@ -63,14 +64,15 @@ def main():
             # Verifying if there is failed commands
             failed_commands = command_manager.get_failed_commands()
             if len(failed_commands) > 0:
-                FAILED_COMMAND_MESSAGE = """{line}
+                FAILED_COMMAND_MESSAGE = dedent("""\
+                {line}
                 {nb_failed} command(s) are in a failed state. They won't be resumed.
                 Failed commands:
                 {failed_commands}
                 The actual errors can be found in the log folder under:
                 {failed_commands_err_file}
-                {line}
-                """
+                {line}\
+                """)
                 print FAILED_COMMAND_MESSAGE.format(
                     line="-" * 84,
                     nb_failed=len(failed_commands),
@@ -110,7 +112,7 @@ def main():
 
     # Launch the jobs
     print "## {nb_commands} command(s) will be executed in {nb_jobs} job(s) ##".format(nb_commands=nb_commands, nb_jobs=len(pbs_filenames))
-    print "Batch UID:\n {batch_uid}".format(batch_uid=jobname)
+    print "Batch UID:\n{batch_uid}".format(batch_uid=jobname)
     if not args.doNotLaunch:
         jobs_id = []
         for pbs_filename in pbs_filenames:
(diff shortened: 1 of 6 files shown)


scripts/smart_dispatch.py
```diff
@@ -3,8 +3,10 @@
 
 import os
 import argparse
+import time as t
 import numpy as np
 from subprocess import check_output
+from textwrap import dedent
```
MarcCote: Good to know that such builtin library exists. :smile:

scripts/smart_dispatch.py
```diff
@@ -59,6 +61,24 @@ def main():
         if args.mode == "launch":
             command_manager.set_commands_to_run(commands)
         else:
+            # Verifying if there is failed commands
```
MarcCote: 'there are'

scripts/smart_dispatch.py
```diff
@@ -59,6 +61,24 @@ def main():
...
+            failed_commands = command_manager.get_failed_commands()
+            if len(failed_commands) > 0:
+                FAILED_COMMAND_MESSAGE = dedent("""\
+                {nb_failed} command(s) are in a failed state. They won't be resumed.
+                Failed commands:
+                {failed_commands}
+                The actual errors can be found in the log folder under:
```
MarcCote: Should we recall where the log folder is?

scripts/smart_dispatch.py
```diff
@@ -88,16 +108,17 @@ def main():
 
     # Launch the jobs
     print "## {nb_commands} command(s) will be executed in {nb_jobs} job(s) ##".format(nb_commands=nb_commands, nb_jobs=len(pbs_filenames))
-    print "Batch UID:\n {batch_uid}".format(batch_uid=jobname)
+    print "Batch UID:\n{batch_uid}".format(batch_uid=jobname)
     if not args.doNotLaunch:
         jobs_id = []
         for pbs_filename in pbs_filenames:
             qsub_output = check_output('{launcher} {pbs_filename}'.format(launcher=LAUNCHER if args.launcher is None else args.launcher, pbs_filename=pbs_filename), shell=True)
-            jobs_id += [qsub_output.rstrip()]
```
MarcCote: Why was there a `rstrip` instead of `strip` in the first place?

scripts/smart_dispatch.py
```diff
@@ -88,16 +108,17 @@ def main():
 
     # Launch the jobs
     print "## {nb_commands} command(s) will be executed in {nb_jobs} job(s) ##".format(nb_commands=nb_commands, nb_jobs=len(pbs_filenames))
-    print "Batch UID:\n {batch_uid}".format(batch_uid=jobname)
+    print "Batch UID:\n{batch_uid}".format(batch_uid=jobname)
```
MarcCote: Héhé, better without the space right :stuck_out_tongue: ?

smartdispatch/command_manager.py
```diff
@@ -40,9 +41,21 @@ def get_nb_commands_to_run(self):
...
+
+    def set_running_command_as_finished(self, command, error_code=0):
+        if error_code == 0:
+            file_name = self._finished_commands_filename
+        else:
+            file_name = self._failed_commands_filename
+
         with utils.open_with_lock(self._running_commands_filename, 'r+') as running_commands_file:
-            with utils.open_with_lock(self._finished_commands_filename, 'a') as finished_commands_file:
+            with utils.open_with_lock(file_name, 'a') as finished_commands_file:
```
MarcCote: Maybe we could use the term 'terminated' instead of 'finished' when want to talk about both finished and failed commands ?

smartdispatch/utils.py
```diff
@@ -5,10 +5,41 @@
...
+    max_len = len(max(splitted_string, key=len))
+    box_line = (u"\u2500" * (max_len + 2))
+
+    out = u"\u250c" + box_line + u"\u2510\n"
+    out += '\n'.join([u"\u2502 {} \u2502".format(line.ljust(max_len)) for line in splitted_string])
+    out += u"\n\u2514" + box_line + u"\u2518"
+    print out
+
+
+def yes_no_prompt(query, default=None):
```
MarcCote: No unit test?

smartdispatch/utils.py
```diff
@@ -5,10 +5,41 @@
 import unicodedata
 import json
 
+from distutils.util import strtobool
 from subprocess import Popen, PIPE
 from contextlib import contextmanager
 
 
+def print_boxed(string):
```
MarcCote: No unit test?

smartdispatch/utils.py
```diff
@@ -5,10 +5,41 @@
...
+    splitted_string = string.split('\n')
+    max_len = len(max(splitted_string, key=len))
```
MarcCote: I found this clearer: `max_len = max(map(len, splitted_string))`

smartdispatch/utils.py
```diff
@@ -5,10 +5,41 @@
...
+    box_line = (u"\u2500" * (max_len + 2))
```
MarcCote: External parentheses are not necessary. You prefer with?

smartdispatch/utils.py
```diff
@@ -5,10 +5,41 @@
...
+
+    out = u"\u250c" + box_line + u"\u2510\n"
+    out += '\n'.join([u"\u2502 {} \u2502".format(line.ljust(max_len)) for line in splitted_string])
+    out += u"\n\u2514" + box_line + u"\u2518"
+    print out
+
+
+def yes_no_prompt(query, default=None):
+    if default is None:
```
MarcCote: What do you think the following instead of the switch?
```python
prompt = {None: " [y/n] ", 'y': " [Y/n] ", 'n': " [y/N] "}

if default not in prompt:
    raise ValueError("Invalid default answer: '{}'".format(default))
```

smartdispatch/utils.py
```diff
@@ -5,10 +5,41 @@
...
+        prompt = " [y/n] "
+    elif default == "y":
+        prompt = " [Y/n] "
+    elif default == "n":
+        prompt = " [y/N] "
+    else:
+        raise ValueError("invalid default answer: '{}'".format(default))
```
MarcCote: Capitalized 'i' => "Invalid..."

smartdispatch/utils.py
```diff
@@ -5,10 +5,41 @@
 import unicodedata
 import json
 
+from distutils.util import strtobool
```
MarcCote: Nice function to be aware :smile: .

smartdispatch/utils.py
```diff
@@ -5,10 +5,41 @@
...
+    elif default == "y":
+        prompt = " [Y/n] "
+    elif default == "n":
+        prompt = " [y/N] "
+    else:
+        raise ValueError("invalid default answer: '{}'".format(default))
+
+    while True:
+        try:
+            answer = raw_input("{0}{1}".format(query, prompt))
```
MarcCote: Would be `prompt[default]` if we use the proposed modification above.

MarcCote:
This is a nice PR we should have had a long time ago. Nothing major to change, some typos here and there and two small unit tests.

scripts/smart_dispatch.py
mgermain: Yeah I though about that. In the case the guy says Yes and want to resume anyway he will see the path to the log folder but in the case he says No you are right it's not appearing anymore. That is partly why I created #81, I think it's fine like that for that pull request and should be cleaned up in a different one because there is more than one Print to add.

scripts/smart_dispatch.py
mgermain: No Idea, I though it was weird. Maybe that at the time I saw that there was only space on the right of the number.

scripts/smart_dispatch.py
mgermain: Yeah I guess ;P

smartdispatch/utils.py
mgermain: We should not test if the print statement is working. I'm not sure how to test that the printout is aesthetically pleasing, the only I could think of would be to verify if the lines are where I put them and I'm really not sure that would be a useful test.

smartdispatch/utils.py
mgermain: We should not test if the print statement is working. I'm not sure how to test that the printout is aesthetically pleasing, the only I could think of would be to verify if the lines are where I put them and I'm really not sure that would be a useful test.

smartdispatch/utils.py
mgermain: Nah I'll remove them I just had an extra + "" at some point.

smartdispatch/utils.py
mgermain: Yeah way better, I was lazy.

smartdispatch/utils.py
MarcCote: I think we should at least test if the function runs correctly (i.e. no exceptions have been raised). Let put a simple test like
```python
def test_printed_box():
    printed_box("")
    printed_box("Some text!")
```

smartdispatch/utils.py
MarcCote: Yeah, you're right.

scripts/smart_dispatch.py
MarcCote: Fine.

PR #85, opened 2015-03-04
Enforce worker use and added a default to the worker pool size

### LINKED ISSUE #35: Add a default # of workers for --pool
Add the possibility to do a ``--pool`` instead of just ``--pool N``.
When you do not specify the N (number of worker in the pool), it is set to the total number of commands or the number of computing unit(only if known queue), whichever the smallest is.

### LINKED ISSUE #79: When generating the job folder name, some characters are removed
When generating the job folder name, some characters are especially removed to make sure the name does not conflict with how the OS handle its files/folders. However, this means that two jobs' folder name will collide if they only differ by some characters that will be removed even though the jobs produce different commands.

For example, the following commands will generate the same folder name:
`smart_dispatch.py -x -q test -C 1 -t 1:00 -p 2 launch script [-option1\ value1 -option1\ value2]`
`smart_dispatch.py -x -q test -C 1 -t 1:00 -p 2 launch script "[-option1\ value1 -option1\ value2]"`

They produce the same folder name: `script__option1_value1__option1_value2/`
but their commands are different (respectively)
```
script -option1
script value1
script -option1
script value2
```
and
```
script -option1 value1
script -option1 value2
```

mgermain:
Close #35
Close #79 The best we'll get until we use UID

### DIFF
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -54,37 +54,41 @@ def main():
         raise ValueError("Unknown subcommand!")
 
     # Pool of workers
-    if args.pool is not None:
-        command_manager = CommandManager(os.path.join(path_job_commands, "commands.txt"))
+    #if args.pool is not None:
+    command_manager = CommandManager(os.path.join(path_job_commands, "commands.txt"))
 
-        # If resume mode, reset running jobs
-        if args.mode == "launch":
-            command_manager.set_commands_to_run(commands)
-        else:
-            # Verifying if there is are failed commands
-            failed_commands = command_manager.get_failed_commands()
-            if len(failed_commands) > 0:
-                FAILED_COMMAND_MESSAGE = dedent("""\
-                {nb_failed} command(s) are in a failed state. They won't be resumed.
-                Failed commands:
-                {failed_commands}
-                The actual errors can be found in the log folder under:
-                {failed_commands_err_file}""")
-                utils.print_boxed(FAILED_COMMAND_MESSAGE.format(
-                    nb_failed=len(failed_commands),
-                    failed_commands=''.join(failed_commands),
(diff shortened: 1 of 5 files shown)


scripts/smart_dispatch.py
```diff
@@ -53,47 +53,43 @@ def main():
...
+        nb_commands = command_manager.get_nb_commands_to_run()
+
+    # If no pool size is specified the number of commands is taken
+    if args.pool is None:
+        args.pool = command_manager.get_nb_commands_to_run()
+
+    # Generating all the worker commands
+    COMMAND_STRING = 'cd "{cwd}"; smart_worker.py "{commands_file}" "{log_folder}" '\
+                     '1>> "{log_folder}/worker/$PBS_JOBID\"\"_worker_{{ID}}.o" '\
+                     '2>> "{log_folder}/worker/$PBS_JOBID\"\"_worker_{{ID}}.e" '
```
MarcCote: The double `{}` is necessary because you call format twice and only specify ID the second time?

scripts/smart_dispatch.py
```diff
@@ -180,6 +176,7 @@ def get_job_folders(jobname):
 
     if not os.path.exists(path_job_logs):
         os.makedirs(path_job_logs)
+        os.makedirs(os.path.join(path_job_logs, "worker"))
```
MarcCote: Cool

smartdispatch/tests/test_smartdispatch.py
```diff
@@ -3,14 +3,15 @@
 
 from nose.tools import assert_true, assert_equal
 from numpy.testing import assert_array_equal
-from datetime import datetime
 from smartdispatch import utils
 
 
 def test_generate_name_from_command():
+    date_lenght = 20
```
MarcCote: `data_lenght` => `data_length`

scripts/smart_dispatch.py
mgermain: yup

Issue #77, opened 2015-01-26
Smarter resume


mgermain:
Considering we have the jobs id logged in jobs_id.txt, when we resume we should verify that they are not still running and if they do ask for a confirmation of the resume.

Also make sure that the jobs a appended properly to jobs_id.txt when we resume.

PR #97, opened 2015-06-18
Resume only pending commands

MarcCote:
I added an option to `smart_dispatch` so we can resume only commands that are pending.

I find it useful in the following context:
Let suppose half of my jobs failed because of easy-to-fix bugs. Then, after I fixed it, I want to relaunch the failed commands (but some commands are already running). Now, I can easily do a `echo failed_commands.txt >> commands.txt` followed by a `smart_dispatch resume JOBIB --onlyPending`.

Without the `--onlyPending` option, the running commands (i.e. found in `running_commands.txt`) would be put along the others commands in `commands.txt` and relaunched!

### DIFF
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -77,7 +77,7 @@ def main():
             if not utils.yes_no_prompt("Do you want to continue?", 'n'):
                 exit()
 
-        if not args.only_pending:
+        if not args.onlyPending:
             command_manager.reset_running_commands()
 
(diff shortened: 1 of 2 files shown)


MarcCote:
This is somewhat related to #77. I would say this PR is a temporary fix until we fully implement #77. I'm afraid that checking if a job is *really* running will highly depend on which cluster we are on.

mgermain:
Do you think it's worth adding a test for this one?

MarcCote:
Definitively. Here you go.

PR #98, opened 2015-06-18
Support both path and Batch UID when resuming jobs.

MarcCote:
Currently we are force to provide the *Batch UID* when resume some jobs with `smart_dispatch`. This PR add also the option of providing the path (i.e. `SMART_DISPATCH_LOGS/{Batch UID}`).

Since I'm using the bash auto-complete to get the *Batch UID*, I find it frustrating to have to remove the `SMART_DISPATCH_LOGS/` part every time.

### DIFF
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -20,7 +20,6 @@ from smartdispatch import utils
 import logging
 import smartdispatch
 
-
 LOGS_FOLDERNAME = "SMART_DISPATCH_LOGS"
 CLUSTER_NAME = utils.detect_cluster()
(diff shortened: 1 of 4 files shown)


MarcCote:
This address the last point of #77.

scripts/smart_dispatch.py
```diff
@@ -166,6 +166,10 @@ def _gen_job_paths(jobname):
 
 
 def get_job_folders(jobname):
+    if os.path.isdir(jobname):
```
mgermain: Looks good!
Adding a small unit test that verify both case might be a good idea.

MarcCote: Added said unit test and use the opportunity to refactor some functions in the script `smart_dispatch.py`.

PR #100, opened 2015-06-25
Use the atomicity of folder creation to simulate a locking mechanism

MarcCote:
This PR changes the `open_with_lock` function so it uses the atomicity of folder creation on clusters filesystems to implement a locking mechanism.

This change is motivated by the fact that the filsesytem used by Helios and Colosse doesn't have the global lock option enabled!

I'm aware it would be better to support both types of lock (i.e. the folder atomicity trick and the python module fcntl) depending on whether the cluster we are on supports or not the global lock option, but I don't know how to do it in a clean way. On the other hand, having only one (universal) type of locking mechanism might be easier to support.

### DIFF
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -28,10 +28,14 @@ LAUNCHER = utils.get_launcher(CLUSTER_NAME)
 
 
 def main():
+    args = parse_arguments()
+
     # Necessary if we want 'logging.info' to appear in stderr.
     logging.root.setLevel(logging.INFO)
 
-    args = parse_arguments()
+    if args.verbose:
+        logging.root.setLevel(logging.DEBUG)
+
     path_smartdispatch_logs = pjoin(os.getcwd(), LOGS_FOLDERNAME)
 
     # Check if RESUME or LAUNCH mode
@@ -152,6 +156,8 @@ def parse_arguments():
     parser.add_argument('-x', '--doNotLaunch', action='store_true', help='Creates the QSUB files without launching them.')
 
     parser.add_argument('-p', '--pool', type=int, help="Number of workers that will be consuming commands. Default: Nb commands")
+
+    parser.add_argument('-v', '--verbose', action='store_true', help='Verbose mode.')
     subparsers = parser.add_subparsers(dest="mode")
 
     launch_parser = subparsers.add_parser('launch', help="Launch jobs.")
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -152,8 +152,6 @@ def parse_arguments():
     parser.add_argument('-x', '--doNotLaunch', action='store_true', help='Creates the QSUB files without launching them.')
 
     parser.add_argument('-p', '--pool', type=int, help="Number of workers that will be consuming commands. Default: Nb commands")
-
-    parser.add_argument('-v', '--verbose', action='store_true', help='Verbose mode.')
     subparsers = parser.add_subparsers(dest="mode")
 
     launch_parser = subparsers.add_parser('launch', help="Launch jobs.")
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -16,6 +16,7 @@ from smartdispatch.queue import Queue
 from smartdispatch.job_generator import job_generator_factory
 from smartdispatch import get_available_queues
 from smartdispatch import utils
+from smartdispatch.filelock import open_with_lock
 
 import logging
 import smartdispatch
@@ -126,7 +127,7 @@ def main():
             qsub_output = check_output('{launcher} {pbs_filename}'.format(launcher=LAUNCHER if args.launcher is None else args.launcher, pbs_filename=pbs_filename), shell=True)
             jobs_id += [qsub_output.strip()]
 
-        with utils.open_with_lock(pjoin(path_job, "jobs_id.txt"), 'a') as jobs_id_file:
+        with open_with_lock(pjoin(path_job, "jobs_id.txt"), 'a') as jobs_id_file:
             jobs_id_file.writelines(t.strftime("## %Y-%m-%d %H:%M:%S ##\n"))
             jobs_id_file.writelines("\n".join(jobs_id) + "\n")
         print "\nJobs id:\n{jobs_id}".format(jobs_id=" ".join(jobs_id))
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -28,14 +28,10 @@ LAUNCHER = utils.get_launcher(CLUSTER_NAME)
 
 
 def main():
-    args = parse_arguments()
-
     # Necessary if we want 'logging.info' to appear in stderr.
     logging.root.setLevel(logging.INFO)
 
-    if args.verbose:
-        logging.root.setLevel(logging.DEBUG)
-
+    args = parse_arguments()
     path_smartdispatch_logs = pjoin(os.getcwd(), LOGS_FOLDERNAME)
 
     # Check if RESUME or LAUNCH mode
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,6 @@ setup(
     license='LICENSE.txt',
     description='A batch job launcher for the Mammouth supercomputer.',
     long_description=open('README.txt').read(),
-    install_requires=[],
+    install_requires=['psutil'],
     package_data={'smartdispatch': ['config/*.json']}
 )
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,6 @@ setup(
     license='LICENSE.txt',
     description='A batch job launcher for the Mammouth supercomputer.',
     long_description=open('README.txt').read(),
-    install_requires=['psutil'],
+    install_requires=['psutil>=1'],
     package_data={'smartdispatch': ['config/*.json']}
 )
--- a/smartdispatch/filelock.py
+++ b/smartdispatch/filelock.py
@@ -1,0 +1,80 @@
+import os
+import time
+import fcntl
+import logging
+
+from contextlib import contextmanager
+
+have_psutil = True
+try:
+    import psutil
+except ImportError:
+    have_psutil = False
+
+# Constants needed for `open_with_dirlock` function.
+MAX_ATTEMPTS = 1000  # This would correspond to be blocked for ~15min.
+TIME_BETWEEN_ATTEMPTS = 1  # In seconds
+
+
+def find_mount_point(path='.'):
+    """ Finds the mount point used to access `path`. """
+    path = os.path.abspath(path)
+    while not os.path.ismount(path):
+        path = os.path.dirname(path)
+
+    return path
+
+
+def get_fs(path='.'):
+    """ Gets info about the filesystem on which `path` lives. """
+    mount = find_mount_point(path)
+
+    for fs in psutil.disk_partitions(True):
+        if fs.mountpoint == mount:
+            return fs
+
+
+@contextmanager
+def open_with_flock(*args, **kwargs):
+    """ Context manager for opening file with an exclusive lock. """
+    f = open(*args, **kwargs)
+    try:
+        fcntl.lockf(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
+    except IOError:
+        logging.info("Can't immediately write-lock the file ({0}), waiting ...".format(f.name))
+        fcntl.lockf(f, fcntl.LOCK_EX)
+
+    yield f
+    fcntl.lockf(f, fcntl.LOCK_UN)
+    f.close()
+
+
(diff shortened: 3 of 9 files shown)


smartdispatch/utils.py
```diff
@@ -9,6 +10,10 @@
 from subprocess import Popen, PIPE
 from contextlib import contextmanager
 
+# Constants needed for `open_with_lock` function
```
mgermain: Should probably be in the function, or named params with default ?

smartdispatch/utils.py
```diff
@@ -84,15 +89,23 @@ def unhexify(match):
 @contextmanager
 def open_with_lock(*args, **kwargs):
```
mgermain: Not sure I like the fact that it's the default. That we don't even try to use an actual lock. Maybe we can find a way to determine this in advance. Maybe looking at the way the current drive is mounted or something then fallback to this?

smartdispatch/utils.py
```diff
@@ -84,15 +89,23 @@ def unhexify(match):
...
-        fcntl.lockf(f, fcntl.LOCK_EX)
-    yield f
-    fcntl.lockf(f, fcntl.LOCK_UN)
-    f.close()
+    dirname = os.path.dirname(args[0])
+    filename = os.path.basename(args[0])
+    lockfile = os.path.join(dirname, "." + filename)
+
+    no_attempt = 0
+    while no_attempt < MAX_ATTEMPTS:
```
mgermain: Active retries with a MAX, this really does not sound safe :P especially for a default behaviour.

smartdispatch/utils.py
mgermain: Also, was it tested manually on problematic cluster?

MarcCote: I was indeed manually tested on "the" problematic cluster :P. I'm also using it a lot and I didn't see anything wrong yet.

smartdispatch/utils.py
MarcCote: The worst that could happen is the job fails after trying for too long. I think this is the best we can assure when using folders as a lock mechanism. Note that if we don't have a MAX and something goes wrong, it might cause an infinite loop, consuming allocation time on the clusters. I think the latter would be worse.

smartdispatch/utils.py
MarcCote: Hmm I get your point. We should have a flag that says "yep, you can use the flock/lockf module". And this flags should be set in some **config file** (we need this so bad) so we don't have to perform the check every time we open a file with a lock.

As for detecting if it is supported, either we hard code it or we find a fancy way of automatically testing it :S !

mgermain: The worker could look if it's a known working filesystem/flags for fcnlt by doing something like this:
```
import os, psutil

def find_mount_point(path):
    path = os.path.abspath(path)
    while not os.path.ismount(path):
        path = os.path.dirname(path)
    return path

mount = find_mount_point(".")

def get_fs(mount):
    for fs in psutil.disk_partitions(True):
        if fs.mountpoint == mount:
            return fs

fs = get_fs(mount)
if fs.fstype == "lustre" and "localflock" not in fs.opts:
    print("YAY can use flock")
```

If not then use the hacky folder thingy.

MarcCote: Interesting.

mgermain: Do you think you cold to it so I merge this one ?

MarcCote:
@mgermain: I put both file locking mechanisms in `smartdispatch/filelock.py`. Upon loading the module, the right one to use is detected. A message is printed if the current cluster does support flock.

setup.py
```diff
@@ -12,6 +12,6 @@
     license='LICENSE.txt',
     description='A batch job launcher for the Mammouth supercomputer.',
     long_description=open('README.txt').read(),
-    install_requires=[],
+    install_requires=['psutil'],
```
mgermain: It's recommended to always put a version number for dependencies.
No need to do an in-depth search to find the oldest compatible version but putting something like psutil>=3.4.2 or whichever lower version you know is working would be good.

smartdispatch/filelock.py
```diff
@@ -0,0 +1,81 @@
...
+import time
+import fcntl
+import logging
+
+from contextlib import contextmanager
+
+have_psutil = True
+try:
+    import psutil
+except ImportError:
```
mgermain: You put psutil in the hard dependencies so there is no need to make this verification as it will always be there.

smartdispatch/utils.py
```diff
@@ -1,13 +1,14 @@
...
 import unicodedata
 import json
 
 from distutils.util import strtobool
 from subprocess import Popen, PIPE
-from contextlib import contextmanager
+
+# Constants needed for `open_with_lock` function
+MAX_ATTEMPTS = 1000
+TIME_BETWEEN_ATTEMPTS = 1  # In seconds
```
mgermain: You forgot to remove this.

tests/test_smart_worker.py
```diff
@@ -106,7 +107,7 @@ def test_lock(self):
         command = ["smart_worker.py", self.command_manager._commands_filename, self.logs_dir]
 
         # Lock the commands file before running 'smart_worker.py'
-        with utils.open_with_lock(self.command_manager._commands_filename, 'r+'):
+        with filelock.open_with_lock(self.command_manager._commands_filename, 'r+'):
```
mgermain: You did `from .filelock import open_with_lock` everywhere but here.

smartdispatch/filelock.py
MarcCote: Yeah, right. Are we certain `psutil` can be installed on every clusters? Also, is there a way to specify *soft* dependencies with pip?

smartdispatch/utils.py
MarcCote: Oupsy.

tests/test_smart_worker.py
MarcCote: I'll change for the sake of consistency. Also, not that I've used relative import since the name of our  library `smartdispatch` conflicts with the module `smartdispatch.py`.

setup.py
MarcCote: Ok, I've tried with psutil==1 and it passed the tests, so I added psutil>=1 as a dependency instead.

smartdispatch/filelock.py
mgermain: Yes, there is a way. https://pythonhosted.org/setuptools/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies but I think this one should be a hard dependency.

I'm pretty sure it's installable everywhere, what is making you doubt ?

scripts/smart_dispatch.py
```diff
@@ -151,6 +152,8 @@ def parse_arguments():
     parser.add_argument('-x', '--doNotLaunch', action='store_true', help='Creates the QSUB files without launching them.')
 
     parser.add_argument('-p', '--pool', type=int, help="Number of workers that will be consuming commands. Default: Nb commands")
+
+    parser.add_argument('-v', '--verbose', action='store_true', help='Verbose mode.')
```
mgermain: What is this? Where is it used?

MarcCote: Ah, it was for logging whether the current cluster supports flock or not. I'll remove it since it is not used for now.

smartdispatch/tests/test_filelock.py
```diff
@@ -0,0 +1,66 @@
...
+import shutil
+
+from subprocess import Popen, PIPE
+from nose.tools import assert_equal, assert_true
+
+from smartdispatch.filelock import open_with_lock, open_with_dirlock, open_with_flock
+from smartdispatch.filelock import find_mount_point, get_fs
+
+
+def _test_open_with_lock(lock_func):
```
mgermain: Maybe it's obvious maybe it's not but what do you think about adding a comment explaining that this test is not complete and is only testing same node lock and not multi-node lock?

smartdispatch/filelock.py
```diff
@@ -0,0 +1,75 @@
...
+        logging.info("Can't immediately write-lock the file ({0}), waiting ...".format(f.name))
+        fcntl.lockf(f, fcntl.LOCK_EX)
+
+    yield f
+    fcntl.lockf(f, fcntl.LOCK_UN)
+    f.close()
+
+
+@contextmanager
+def open_with_dirlock(*args, **kwargs):
```
mgermain: I manually tested this on helios on 2 different nodes and it's working perfectly :)

smartdispatch/filelock.py
```diff
@@ -0,0 +1,75 @@
...
+            break
+        except OSError:
+            logging.info("Can't immediately write-lock the file ({0}), retrying in {1} sec. ...".format(filename, TIME_BETWEEN_ATTEMPTS))
+            time.sleep(TIME_BETWEEN_ATTEMPTS)
+            no_attempt += 1
+
+
+# Determine if we can rely on the fcntl module for locking files on the cluster.
+# Otherwise, fallback on using the directory creation atomicity as a locking mechanism.
+open_with_lock = open_with_dirlock
```
mgermain: Why not in the else?

smartdispatch/filelock.py
```diff
@@ -0,0 +1,75 @@
...
+fs = get_fs('.')
+if fs.fstype == "lustre" and "localflock" not in fs.opts:
+    print("Cluster supports flock.")
```
mgermain: This should be a log.warning when it does NOT support the standard fcntl.

smartdispatch/filelock.py
```diff
@@ -0,0 +1,75 @@
...
+            logging.info("Can't immediately write-lock the file ({0}), retrying in {1} sec. ...".format(filename, TIME_BETWEEN_ATTEMPTS))
+            time.sleep(TIME_BETWEEN_ATTEMPTS)
+            no_attempt += 1
+
+
+# Determine if we can rely on the fcntl module for locking files on the cluster.
+# Otherwise, fallback on using the directory creation atomicity as a locking mechanism.
+open_with_lock = open_with_dirlock
+fs = get_fs('.')
+if fs.fstype == "lustre" and "localflock" not in fs.opts:
```
mgermain: Did guilimin support globalflock? If so, we should add gpfs to the whitelist of fs.fstype.
Maybe all in a function _fs_support_flock(fs) ?

For safety, we should probably also add "flock" in fs.opts for lustre.

mgermain: Just tested it works on guilimin.

smartdispatch/filelock.py
MarcCote: No particular reason. Personal taste. I found that it reads better.

smartdispatch/filelock.py
MarcCote: Yeah, ok I'll change the logic so that we print the message, only if fcntl is not supported.

The problem with `logging` is that this particular line is executed when the module is loaded which happens during the imports of `scripts/smart_dispatch.py`. This is problematic because the logging level is not set yet. We set it in the `main` function of the script. In others words, we'll never see the message :-/ .

smartdispatch/filelock.py
MarcCote: So nothing to do? Not sure I understand.

MarcCote: What do you have in mind? What about Mammoth, is it lustre also? What fs.opts should I check for?
```python
def _fs_support_flock(fs):
    if fs.fstype == "lustre":
        return "localflock" not in fs.opts

    elif fs.fstype == "gpfs":
        return ?

    return False  # We don't know.
```

mgermain: def _fs_support_flock(fs):
    if fs.fstype == "lustre":
        return ("localflock" not in fs.opts) and ("flock" in fs.opts)
    elif fs.fstype == "gpfs":
        return True
    else:
        return False  # We don't know.

smartdispatch/filelock.py
mgermain: Can we set it before the imports? Is that a valid thing?
There is probably something we are doing wrong if we can't have that.

smartdispatch/filelock.py
mgermain: Oh and yes mammouth is lustre.
Helios = lustre with localflock flag
Mammouth = lustre with flock flag

fcntl is working on mammouth but not on helios

smartdispatch/filelock.py
MarcCote: I'll leave it as a print for now. Will open an issue the rethink the logging.

smartdispatch/filelock.py
MarcCote: Cool

smartdispatch/tests/test_filelock.py
MarcCote: Done

PR #101, opened 2015-06-26
Save the command line in a file to keep an history

MarcCote:
It is often useful to check what was exactly the command I used to launch some experiments with `smart_dispatch`. This PR saves the command in a file in the job folder.

### DIFF
--- a/scripts/smart_dispatch.py
+++ b/scripts/smart_dispatch.py
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
 
+import re
 import os
+import sys
 import argparse
 import time as t
 import numpy as np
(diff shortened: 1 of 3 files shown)


scripts/smart_dispatch.py
```diff
@@ -102,6 +104,14 @@ def main():
     job_generator = job_generator_factory(queue, commands, command_params, CLUSTER_NAME, path_job)
     pbs_filenames = job_generator.write_pbs_files(path_job_commands)
 
+    # Keep an history of command lines i.e. smart_dispatch.py commands.
```
mgermain: I think this should be called at the beginning of the main before the parse_arguments().
Also, this could be a function named `log_command_line` or something similar.

scripts/smart_dispatch.py
```diff
@@ -102,6 +104,14 @@ def main():
...
+    with utils.open_with_lock(os.path.join(path_job, "command_line_history.txt"), 'a') as command_line_history_file:
```
mgermain: command_line.log

scripts/smart_dispatch.py
```diff
@@ -102,6 +104,14 @@ def main():
...
+        command_line_history_file.write(t.strftime("## %Y-%m-%d %H:%M:%S ##\n"))
+        command_line = " ".join(sys.argv)
+        command_line = command_line.replace('"', r'\"')  # Make sure we can paste the command line as-is
+        command_line = re.sub(r'(\[)([^\[\]]*\\ [^\[\]]*)(\])', r'"\1\2\3"', command_line)  # Make sure we can paste the command line as-is
```
mgermain: Add a unit test to be sure the command is reconstructed properly?

scripts/smart_dispatch.py
MarcCote: I added a function in `smartdispatch/smartdispatch.py`.

scripts/smart_dispatch.py
MarcCote: Yep, I also think that's a better name

scripts/smart_dispatch.py
MarcCote: Done, see `I added a function in `smartdispatch/tests/test_smartdispatch.py`.`

PR #104, opened 2016-02-08
Make sure commands output/error are appended to the log.

MarcCote:
Right now, when using `smart_dispatch.py resume ...` the output and error logs are overwritten. This PR makes sure we append the content to the log files accordingly. It also append the data and the time of the (re)execution of the command.


scripts/smart_worker.py
```diff
@@ -44,11 +45,22 @@ def main():
...
+
+        with open(stdout_filename, 'a') as stdout_file:
+            with open(stderr_filename, 'a') as stderr_file:
+                if stdout_already_exists:
+                    stdout_file.write('\n')
+
+                if stderr_already_exists:
+                    stderr_file.write('\n')
+
+                stdout_file.write(t.strftime("## %Y-%m-%d %H:%M:%S ##\n"))
```
mgermain: You could take the timestamp only once, so the resume time will be the same in both log file.
And we could make de difference between ## Started on: {date} ## and ## Resumed on : {date} ##

Also considering this file is a "user" file maybe we should do ## SMART_DISPATCH Started on: {date} ## so there is no confusion about where the print comes from?

MarcCote: I agree with all this, nice ideas.

PR #105, opened 2016-02-08
[FIX] use OrderedDict in PBS. [PEP8] replaced tabs with spaces.

MarcCote:
Sometime the test in `test_pbs.py` were falling because we use `dict` to store options and resources. So writing the PBS file, we were iterating over the items of those dictionaries which is nondeterminist (depends on the hash of the keys). This PR fixes this by using OrderedDict instead.

Also, fixed some minor PEP8 errors.

### DIFF
--- a/smartdispatch/tests/test_pbs.py
+++ b/smartdispatch/tests/test_pbs.py
@@ -72,8 +72,8 @@ class TestPBS(unittest.TestCase):
     def test_str(self):
         # Create simple PBS file
         expected = """#!/bin/bash
-#PBS -V
 #PBS -q qtest@mp2
+#PBS -V
 #PBS -l walltime=01:00:00
 
 # Modules #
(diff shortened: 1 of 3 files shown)


tests/test_smart_dispatch.py
```diff
@@ -62,7 +62,6 @@ def test_main_resume(self):
 
         # Test when batch_uid is a path instead of a jobname.
         # Setup
-        call(self.launch_command, shell=True)
```
MarcCote: This one was not necessary.

smartdispatch/tests/test_pbs.py
```diff
@@ -90,11 +90,11 @@ def test_str(self):
         modules = ["CUDA_Toolkit/6.0", "python2.7"]
 
         expected = """#!/bin/bash
```
MarcCote: Order of the options and resources have changed.

smartdispatch/tests/test_pbs.py
```diff
@@ -72,8 +72,8 @@ def test_add_commands(self):
     def test_str(self):
         # Create simple PBS file
         expected = """#!/bin/bash
```
MarcCote: Order of the options and resources have changed.

PR #106, opened 2016-02-08
Add Python 3 support

### LINKED ISSUE #89: Update shebang to python2
Investigate what would be the best way to make sure the script always run on python2.

MarcCote:
It's all in the title. :D

Maybe @ASalvail could you have a quick look at this one whenever you have time.
Edit: closes #89

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 python:
   - "2.7"
-  #- "3.4"
+  - "3.4"
 
 install:
   - pip install coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: python
 python:
   - "2.7"
   - "3.4"
+  - "3.5"
 
 install:
   - pip install coveralls
--- a/smartdispatch/argument_template.py
+++ b/smartdispatch/argument_template.py
@@ -35,7 +35,7 @@ class RangeArgumentTemplate(ArgumentTemplate):
         start = int(groups[0])
         end = int(groups[1])
         step = 1 if groups[2] is None else int(groups[2])
-        return map(str, range(start, end, step))
+        return list(map(str, range(start, end, step)))
 
 
 argument_templates = build_argument_templates_dictionnary()
--- a/smartdispatch/argument_template.py
+++ b/smartdispatch/argument_template.py
@@ -35,7 +35,7 @@ class RangeArgumentTemplate(ArgumentTemplate):
         start = int(groups[0])
         end = int(groups[1])
         step = 1 if groups[2] is None else int(groups[2])
-        return list(map(str, range(start, end, step)))
+        return [str(i) for i in range(start, end, step)]
 
 
 argument_templates = build_argument_templates_dictionnary()
--- a/smartdispatch/smartdispatch.py
+++ b/smartdispatch/smartdispatch.py
@@ -97,7 +97,7 @@ def unfold_command(command):
         pos = match.end()
 
     arguments.append([text[pos:]])  # Add remaining unfolded arguments
-    arguments = [map(utils.decode_escaped_characters, argvalues) for argvalues in arguments]
+    arguments = [list(map(utils.decode_escaped_characters, argvalues)) for argvalues in arguments]
     return ["".join(argvalues) for argvalues in itertools.product(*arguments)]
 
 
--- a/smartdispatch/smartdispatch.py
+++ b/smartdispatch/smartdispatch.py
@@ -97,7 +97,7 @@ def unfold_command(command):
         pos = match.end()
 
     arguments.append([text[pos:]])  # Add remaining unfolded arguments
-    arguments = [list(map(utils.decode_escaped_characters, argvalues)) for argvalues in arguments]
+    arguments = [[utils.decode_escaped_characters(v) for v in argvalues] for argvalues in arguments]
     return ["".join(argvalues) for argvalues in itertools.product(*arguments)]
 
 
--- a/smartdispatch/utils.py
+++ b/smartdispatch/utils.py
@@ -57,11 +57,13 @@ def slugify(value):
     ---------
     https://github.com/django/django/blob/1.7c3/django/utils/text.py#L436
     """
+    # Convert `value` to Unicode so we can slugify it using the unicodedata module.
     try:
         value = unicode(value, "UTF-8")
     except NameError:
-        pass  # In Python 3 all strings are already unicode.
+        pass  # In Python 3, all strings are already stored as Unicode.
 
+    # Replace all compatibility characters with their equivalents.
     value = unicodedata.normalize('NFKD', value).encode('ascii', 'ignore').decode('ascii')
     value = re.sub('[^\w\s-]', '', value).strip().lower()
     return str(re.sub('[-\s]+', '_', value))
--- a/smartdispatch/utils.py
+++ b/smartdispatch/utils.py
@@ -1,6 +1,7 @@
 from __future__ import print_function
 
 import re
+import binascii
 import hashlib
 import unicodedata
 import json
@@ -37,13 +38,13 @@ def yes_no_prompt(query, default=None):
 
 def chunks(sequence, n):
     """ Yield successive n-sized chunks from sequence. """
-    for i in xrange(0, len(sequence), n):
+    for i in range(0, len(sequence), n):
         yield sequence[i:i + n]
 
 
 def generate_uid_from_string(value):
     """ Create unique identifier from a string. """
-    return hashlib.sha256(value).hexdigest()
+    return hashlib.sha256(value.encode()).hexdigest()
 
 
 def slugify(value):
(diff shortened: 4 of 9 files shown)


mgermain:
This makes #89 Invalid.

scripts/smart_dispatch.py
```diff
@@ -145,7 +150,7 @@ def parse_arguments():
     parser.add_argument('-c', '--coresPerCommand', type=int, required=False, help='How many cores a command needs.', default=1)
     parser.add_argument('-g', '--gpusPerCommand', type=int, required=False, help='How many gpus a command needs.', default=1)
     # parser.add_argument('-m', '--memPerCommand', type=float, required=False, help='How much memory a command needs (in Gb).')
-    parser.add_argument('-f', '--commandsFile', type=file, required=False, help='File containing commands to launch. Each command must be on a seperate line. (Replaces commandAndOptions)')
+    parser.add_argument('-f', '--commandsFile', type=str, required=False, help='File containing commands to launch. Each command must be on a seperate line. (Replaces commandAndOptions)')
```
mgermain: Why change file to str ?

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: python
 python:
   - "2.7"
-  #- "3.4"
+  - "3.4"
```
ASalvail: Come on, even Ubuntu is on 3.5 by now!

smartdispatch/argument_template.py
```diff
@@ -35,7 +35,7 @@ def unfold(self, match):
         start = int(groups[0])
         end = int(groups[1])
         step = 1 if groups[2] is None else int(groups[2])
-        return map(str, range(start, end, step))
+        return list(map(str, range(start, end, step)))
```
ASalvail: Should be ``return [str(x) for x in range(start, end, step)]``

smartdispatch/smartdispatch.py
```diff
@@ -96,7 +96,7 @@ def unfold_command(command):
         pos = match.end()
 
     arguments.append([text[pos:]])  # Add remaining unfolded arguments
-    arguments = [map(utils.decode_escaped_characters, argvalues) for argvalues in arguments]
+    arguments = [list(map(utils.decode_escaped_characters, argvalues)) for argvalues in arguments]
```
ASalvail: ``arguments = [[utils.decode_escaped_characters(val) for val in argvalues] for argvalues in arguments]``

smartdispatch/tests/test_job_generator.py
```diff
@@ -22,7 +22,7 @@ def setUp(self):
         self.mem_per_node = 32
         self.modules = ["cuda", "python"]
 
-        self.queue = Queue(self.name, self.cluster_name, self.walltime, self.cores, 0, self.modules)
+        self.queue = Queue(self.name, self.cluster_name, self.walltime, self.cores, 0, self.mem_per_node, self.modules)
```
ASalvail: There is no changes to the `Queue` constructor arguments. Either that's a bug fix or an error.

ASalvail:
Not familiar with the code base, but with the comments addressed, the changes look fine to me.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: python
```
mgermain: Wh should add unit test in python 3

smartdispatch/argument_template.py
MarcCote: Yep, way cleaner! Thx.

smartdispatch/smartdispatch.py
MarcCote: Thx.

smartdispatch/tests/test_job_generator.py
MarcCote: Good catch! That modification was introduced by another PR. I'll rebase from master so hopefully this changes should disappear.

smartdispatch/utils.py
```diff
@@ -57,7 +60,12 @@ def slugify(value):
     ---------
     https://github.com/django/django/blob/1.7c3/django/utils/text.py#L436
     """
-    value = unicodedata.normalize('NFKD', unicode(value, "UTF-8")).encode('ascii', 'ignore').decode('ascii')
+    try:
+        value = unicode(value, "UTF-8")
+    except NameError:
+        pass  # In Python 3 all strings are already unicode.
+
+    value = unicodedata.normalize('NFKD', value).encode('ascii', 'ignore').decode('ascii')
```
MarcCote: Sure I'll add one. I'm following the "It's easier to ask forgiveness than permission" principle here, it is not really a hack.

tests/test_smart_worker.py
```diff
@@ -22,7 +22,7 @@ def setUp(self):
         self.command_manager = CommandManager(os.path.join(self._commands_dir, "commands.txt"))
         self.command_manager.set_commands_to_run(self.commands)
 
-        self.commands_uid = map(utils.generate_uid_from_string, self.commands)
+        self.commands_uid = list(map(utils.generate_uid_from_string, self.commands))
```
MarcCote: Got it.

.travis.yml
MarcCote: Already done :-), check below.

.travis.yml
MarcCote: Yep, I added 3.5.

MarcCote:
Thank you @ASalvail for your comments.

MarcCote:
@mgermain I added a unit test for the --commandsFile option. Make sure you check https://github.com/SMART-Lab/smartdispatch/pull/106/files#diff-f3e1285c65572b36c9b0f21b0734c260

.travis.yml
mgermain: There is no need to ask for more than we need.
On top of that, this is a cluster tool and more often than not clusters have old version of everything.

MarcCote: smartdispatch can also be used on our own computer (e.g. use the useful expansion of folded argument to generate a commands file :-P) and I have a virtual env with Python 3.5.

Also, testing on Python 3.5 came for free with Travis, why not use take advantage of it?

mgermain: Oh! You test it on both, nevermind I thought you removed 3.4. My bad

smartdispatch/argument_template.py
```diff
@@ -35,7 +35,7 @@ def unfold(self, match):
         start = int(groups[0])
         end = int(groups[1])
         step = 1 if groups[2] is None else int(groups[2])
-        return map(str, range(start, end, step))
+        return [str(i) for i in range(start, end, step)]
```
mgermain: Were we really using map everywhere for no reasons ?

smartdispatch/tests/test_job_generator.py
mgermain: I think we were not passing the mem_per_node for a reason. PBS generator is not using it ?

smartdispatch/utils.py
```diff
@@ -38,13 +41,13 @@ def yes_no_prompt(query, default=None):
 
 def chunks(sequence, n):
     """ Yield successive n-sized chunks from sequence. """
-    for i in xrange(0, len(sequence), n):
+    for i in range(0, len(sequence), n):
```
mgermain: We should probably use the `six` package for translation from python 2 to 3.
Especially in this case because in python2 the range will be super slow. On the other hand we probably don't need it to be that fast.

smartdispatch/utils.py
```diff
@@ -57,15 +60,23 @@ def slugify(value):
     ---------
     https://github.com/django/django/blob/1.7c3/django/utils/text.py#L436
     """
-    value = unicodedata.normalize('NFKD', unicode(value, "UTF-8")).encode('ascii', 'ignore').decode('ascii')
+    # Convert `value` to Unicode so we can slugify it using the unicodedata module.
+    try:
+        value = unicode(value, "UTF-8")
+    except NameError:
```
mgermain: There is probably a cleaner way to manage this. Maybe with six ?

tests/test_smart_dispatch.py
```diff
@@ -15,15 +15,19 @@ def setUp(self):
...
+                         "echo 2 6 9", "echo 2 6 0", "echo 2 7 9", "echo 2 7 0", "echo 2 8 9", "echo 2 8 0",
+                         "echo 3 6 9", "echo 3 6 0", "echo 3 7 9", "echo 3 7 0", "echo 3 8 9", "echo 3 8 0",
+                         "echo 4 6 9", "echo 4 6 0", "echo 4 7 9", "echo 4 7 0", "echo 4 8 9", "echo 4 8 0"]
+        self.nb_commands = len(self.commands)
 
-        smart_dispatch_command = 'smart_dispatch.py -C 1 -q test -t 5:00 -x {0}'
-        self.launch_command = smart_dispatch_command.format('launch ' + self.commands)
-        self.resume_command = smart_dispatch_command.format('resume {0}')
+        self.smart_dispatch_command = 'smart_dispatch.py -C 1 -q test -t 5:00 -x'
+        self.launch_command = self.smart_dispatch_command + " launch " + self.folded_commands
```
mgermain: Why remove the format?

smartdispatch/utils.py
MarcCote: `six` module doesn't seem to have a `unicode` function, maybe it is another name.

smartdispatch/argument_template.py
MarcCote: Yes and no. In Python 3, map returns a generator instead of a list. What we want here is a list, so we are better off using list comprehension instead.

smartdispatch/tests/test_job_generator.py
MarcCote: This is a fixed that should be in a separate PR! See #114 instead.

smartdispatch/utils.py
MarcCote: If we can avoid using `six` I think it is for the best. I say we don't really about speed here. Also, I don't think `range` in Python 2 is that slow.

tests/test_smart_dispatch.py
MarcCote: Because `self.smart_dispatch_command` doesn't have a '{0}' anymore. This is because I'm reusing that variable in the test for `--commandsFile` where I needed to append `-f commands.txt` to the launch command.

MarcCote:
See PR #161

PR #115, opened 2016-04-10
Added the ability to get the RAPID from a file and modified Helios generator

mgermain:


smartdispatch/job_generator.py
```diff
@@ -136,15 +150,10 @@ def generate_pbs(self):
 class HeliosJobGenerator(JobGenerator):
 
     def generate_pbs(self):
-        pbs_list = self.generate_pbs_with_account_name_from_env('RAP')
+        pbs_list = self.generate_pbs_with_account_name_from_file("{}/.default_rap".format(os.environ['HOME']))
 
         for pbs in pbs_list:
-            # Remove forbidden ppn option. Default is 5 cores per 2 gpu.
+            # Remove forbidden ppn option. Default is 2 cores per gpu.
```
MarcCote: Why did we force having an even number of gpus before?

mgermain: It was a restriction of helios that is no longer true.

smartdispatch/job_generator.py
```diff
@@ -136,15 +150,10 @@ def generate_pbs(self):
 class HeliosJobGenerator(JobGenerator):
 
     def generate_pbs(self):
-        pbs_list = self.generate_pbs_with_account_name_from_env('RAP')
+        pbs_list = self.generate_pbs_with_account_name_from_file("{}/.default_rap".format(os.environ['HOME']))
```
MarcCote: It is better to use `os.path.join(os.environ['HOME'], ".default_rap")`.

smartdispatch/tests/test_utils.py
```diff
@@ -10,8 +10,8 @@
 class PrintBoxedTests(unittest.TestCase):
 
     def setUp(self):
-        self.empty = ''
-        self.text = "This is weird test for a visual thing.\nWell maybe it's fine to test it's working."
+        self.empty = u""
```
MarcCote: Unless we specifically need unicode text, I don't think we should force it.

smartdispatch/job_generator.py
mgermain: True!

smartdispatch/job_generator.py
```diff
@@ -100,6 +100,20 @@ def generate_pbs_with_account_name_from_env(self, environment_variable_name):
 
         return pbs_list
 
+    def generate_pbs_with_account_name_from_file(self, rapid_filename):
+        pbs_list = JobGenerator.generate_pbs(self)
```
MarcCote: Why not using `self.generate_pbs()` ?

smartdispatch/job_generator.py
```diff
@@ -100,6 +100,20 @@ def generate_pbs_with_account_name_from_env(self, environment_variable_name):
...
+
+        if not os.path.isfile(rapid_filename):
+            raise ValueError("Account name file {} does not exist. Please, provide your account name!".format(rapid_filename))
```
MarcCote: Should we mention how to provide it (what option to use)?

smartdispatch/job_generator.py
mgermain: Because it would to a recursive call to the generate_pbs() of the child class and explode.

smartdispatch/tests/test_job_generator.py
```diff
@@ -134,28 +134,33 @@ def setUp(self):
         self.commands = ["echo 1", "echo 2", "echo 3", "echo 4"]
         self.queue = Queue("gpu_8", "helios")
 
-        self.env_val = 'RAP'
+        self._home_backup = os.environ['HOME']
+        os.environ['HOME'] = os.path.abspath('.')
```
MarcCote: To be safe we should run the tests in a temp folder like we did here:
https://github.com/SMART-Lab/smartdispatch/blob/master/tests/test_smart_dispatch.py#L15
and change directory using:
https://github.com/SMART-Lab/smartdispatch/blob/master/tests/test_smart_dispatch.py#L29

and making sure in the tearDown we do a clean up:
https://github.com/SMART-Lab/smartdispatch/blob/master/tests/test_smart_dispatch.py#L32

smartdispatch/tests/test_job_generator.py
```diff
@@ -134,28 +134,33 @@ def setUp(self):
...
 
-        self.bak_env_home_group = os.environ.get(self.env_val)
-        if self.bak_env_home_group is not None:
-            del os.environ[self.env_val]
-        os.environ[self.env_val] = "/rap/group/"
+        self.rap_filename = "{}/.default_rap".format(os.environ['HOME'])
+        if os.path.isfile(self.rap_filename):
+            raise Exception("Test fail: {} should not be there.".format(self.rap_filename))
```
MarcCote: This wouln't be necessary if we you use a temp folder to run the tests.

smartdispatch/job_generator.py
MarcCote: Cool

smartdispatch/job_generator.py
mgermain: Good point but I think is kinda explicit that they need to create the file (which should be there, cluster default). How would you phrase it?

smartdispatch/job_generator.py
MarcCote: Right. We should use `super(JobGenerator, self).generate_pbs()` though.

smartdispatch/tests/test_job_generator.py
```diff
@@ -134,28 +134,33 @@ def setUp(self):
...
 
-        self.env_val = 'RAP'
+        self._home_backup = os.environ['HOME']
+        os.environ['HOME'] = os.path.abspath('.')
 
-        self.bak_env_home_group = os.environ.get(self.env_val)
-        if self.bak_env_home_group is not None:
-            del os.environ[self.env_val]
-        os.environ[self.env_val] = "/rap/group/"
+        self.rap_filename = "{}/.default_rap".format(os.environ['HOME'])
```
MarcCote: Same thing here, use `os.path.join`.

smartdispatch/job_generator.py
mgermain: Is that even compatible with python3, I think it's only `super()` ?
Worst case, `super(self.__class__, self).generate_pbs()`

smartdispatch/tests/test_job_generator.py
mgermain: Good point, changing.

smartdispatch/tests/test_job_generator.py
mgermain: Yup

smartdispatch/tests/test_job_generator.py
mgermain: True but still it not totally wortless.

smartdispatch/tests/test_utils.py
mgermain: meh, not sure about that, removing it

smartdispatch/tests/test_job_generator.py
mgermain: I don't think that the changedir is useful in this case.

smartdispatch/job_generator.py
MarcCote: Yes, you can provide the arguments for `super` in Python 3.
You might be right about using `super(self.__class__, self).generate_pbs()` though.

smartdispatch/job_generator.py
MarcCote: I thought we could provide an argument to the script `smart_dispatch.py`

smartdispatch/tests/test_job_generator.py
MarcCote: You are right because of the modification to `os.environ['HOME']`.

smartdispatch/tests/test_job_generator.py
MarcCote: The temp folder will be empty. Maybe I'm missing something but that line will never be executed.

smartdispatch/tests/test_utils.py
MarcCote: thx

smartdispatch/job_generator.py
mgermain: `smart_dispatch.py` never had a account name flag. The reason we never did that is that not all cluster require one.

MarcCote: I see. Supporting adding any PBS flags (not only resources) would be useful for that I guess :P. Regarding the error message, what about: `"Cannot find account name (RAP) in {}".format(rapid_filename)`

smartdispatch/job_generator.py
mgermain: Ok the internet seems to say that super(self.__class__, self).function() is a bad idea as it will hit infinit recursion some times. <ParentClassname>.function() seems to be fine but not recommended because it's a weird syntax.

Python3 is super() but also accept super(<classname>, self).function().

I guess I'll switch everything to super(<classname>, self).function() :S

Issue #88, opened 2015-03-16
Specifiy dependencies


mgermain:
Mage sure that all the dependencies are in the setpu.py (see http://www.scotttorborg.com/python-packaging/dependencies.html) to ensure a clean installation with pip.

MarcCote:
I think we are already doing that, no?

mgermain:
No, we have at least a dependency on Numpy in `smart_dispatch.py` that is not in the `setup.py` and probably others.

mgermain:
We only had numpy and only for a np.inf so we just removed it in PR #117

PR #117, opened 2016-04-11
Renamed script smart_dispatch.py to smart-dispatch

MarcCote:
This PR changes the name of the script `smart_dispatch.py` to a more friendly one: `smart-dispatch`.
For the sake of consistency, this PR also changes the name of the library: `Smart-Dispatch` to `smart-dispatch`.

This PR moves the script `scripts/smart_worker.py` to `smardispatch/workers/base_worker.py" and won't be visible on the PATH anymore. 

This PR also increases the version number and `numpy` as one of the dependencies.

**Edit**
Passing unit tests and simple `smart-dispatch ... launch echo [1 2 3]`
- [x] Mammoth (mp2) 
- [x] guillmin
- [x] helios

**Edit2**
Warning! You need to uninstall your current installation of Smart-Dispatch first. It seems that pip only overwrites old scripts by the new ones. If you change/remove some scripts you have to do the clean up yourself. It is the case for `smart_dispatch.py` which has been renamed to `smart-dispatch` and `smart_worker.py` which has been removed.


setup.py
```diff
@@ -2,16 +2,17 @@
 from distutils.core import setup
 
 setup(
-    name='Smart-Dispatch',
-    version='1.2.1',
+    name='smart-dispatch',
+    version='1.3.1',
```
mgermain: 1.3.0 ?

MarcCote: oups

setup.py
```diff
@@ -2,16 +2,17 @@
...
     author_email='smart-udes-dev@googlegroups.com',
-    packages=['smartdispatch'],
-    scripts=['scripts/smart_dispatch.py', 'scripts/smart_worker.py'],
+    packages=['smartdispatch',
+              'smartdispatch/workers'],
+    scripts=['scripts/smart-dispatch'],
     url='https://github.com/SMART-Lab/smartdispatch',
     license='LICENSE.txt',
-    description='A batch job launcher for the Mammouth supercomputer.',
+    description='A batch job launcher for computer clusters.',
```
mgermain: Weird that we never updated that.
The current tagline is : An easy to use job launcher for supercomputers with PBS compatible job manager.

mgermain:
You should update the readme also.

scripts/smart-dispatch
```diff
@@ -101,10 +101,11 @@ def main():
         args.pool = command_manager.get_nb_commands_to_run()
```
mgermain: Since we are playing in smart_dispatch maybe you should do this tiny one at the same time?https://github.com/SMART-Lab/smartdispatch/issues/89

setup.py
```diff
@@ -2,16 +2,17 @@
...
     long_description=open('README.txt').read(),
-    install_requires=['psutil>=1'],
+    install_requires=['psutil>=1', 'numpy>=1.7'],
```
mgermain: Cool that you are fixing this but, are you sure there is nothing more and it fixes #88 ?

mgermain:
In `base_worker.py` we should probably change `log_datetime = t.strftime("## SMART_DISPATCH` to  `log_datetime = t.strftime("## SMART-DISPATCH` for consistency? Keeping the caps.

setup.py
MarcCote: I've looked at every import of every file. I saw no special library. And we have a numpy dependcy only because we use `np.inf` somewhere. Maybe we could replace it with `math.inf`?

scripts/smart-dispatch
MarcCote: The answer is to support Python 3.

setup.py
MarcCote: Right.

MarcCote:
I changed for "SMART-DISPATCH" when we write in the logs. I also updated the README.md

scripts/smart-dispatch
MarcCote: Okay, for now, I'm updating the shebang to python2.

setup.py
mgermain: You are totally right, we should not have a dependency on numpy one thing that can easily be done without numpy.

That said, `math.inf` is only in python3 (maybe in six?) I think way to do it in python2 is float('inf')

MarcCote: Yep, that's what I did.

PR #121, opened 2016-04-14
Job generator refactoring

mgermain:
Refactored job_generator.py to allow an easier incorporation of the --pbsFlags option and cleaned up some tests.

### DIFF
--- a/smartdispatch/job_generator.py
+++ b/smartdispatch/job_generator.py
@@ -42,7 +42,16 @@ class JobGenerator(object):
         self.nb_gpus_per_command = command_params.get('nb_gpus_per_command', 1)
         #self.mem_per_command = command_params.get('mem_per_command', 0.0)
 
-    def generate_pbs(self):
+        self.pbs_list = self._generate_base_pbs()
+        if type(self) is not JobGenerator:
+            self._add_cluster_specific_rules()
+
+    def add_pbs_flags(self, flags):
+        for flag in flags:
+            if flag.startswith('-l'):
+                flag[2:].trim()
+
+    def _generate_base_pbs(self):
         """ Generates PBS files allowing the execution of every commands on the given queue. """
         nb_commands_per_node = self.queue.nb_cores_per_node // self.nb_cores_per_command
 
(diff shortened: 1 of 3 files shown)


smartdispatch/job_generator.py
```diff
@@ -42,7 +42,16 @@ def __init__(self, queue, commands, command_params={}, base_path="./"):
         self.nb_gpus_per_command = command_params.get('nb_gpus_per_command', 1)
         #self.mem_per_command = command_params.get('mem_per_command', 0.0)
 
-    def generate_pbs(self):
+        self.pbs_list = self._generate_base_pbs()
+        if type(self) is not JobGenerator:
```
MarcCote: Instead of doing this check, we could just call: `self._add_cluster_specific_rules()` and add `_add_cluster_specific_rules` as a virtual method of `JobGenerator` that subclass could override if needed.

smartdispatch/job_generator.py
```diff
@@ -42,7 +42,16 @@ def __init__(self, queue, commands, command_params={}, base_path="./"):
...
+            self._add_cluster_specific_rules()
+
+    def add_pbs_flags(self, flags):
+        for flag in flags:
```
MarcCote: Can you add a docstring explaining  what is `flags` exactly.

smartdispatch/job_generator.py
```diff
@@ -79,79 +88,63 @@ def write_pbs_files(self, pbs_dir="./"):
         pbs_dir : str
             folder where to save pbs files
         """
-        pbs_list = self.generate_pbs()
+        #self.pbs_list = self._generate_base_pbs()
```
MarcCote: Unused?

smartdispatch/tests/test_job_generator.py
```diff
@@ -10,7 +9,7 @@
 from smartdispatch.job_generator import GuilliminJobGenerator, MammouthJobGenerator
 
 
-class TestJobGenerator(unittest.TestCase):
+class TestJobGenerator(object):
```
MarcCote: Are `setUp` and `tearDown` still get called? If so, what was the utility of `unittest.TestCase`?!

smartdispatch/tests/test_job_generator.py
```diff
@@ -154,37 +153,39 @@ def tearDown(self):
...
 
     def test_generate_pbs_valid_group(self):
-        pbs = self.job_generator.generate_pbs()[0]
+        pbs = self.job_generator.pbs_list[0]
 
         assert_equal(pbs.options['-A'], self.rapid)
 
     def test_generate_pbs_ppn_is_absent(self):
-        assert_false("ppn=" in self.job_generator.generate_pbs()[0].__str__())
+        assert_false("ppn=" in self.job_generator.pbs_list[0].__str__())
```
MarcCote: I'm thinking out loud here, why are we calling `__str__` directly instead of using `str(self.job_generator.pbs_list[0])`?

smartdispatch/tests/test_job_generator.py
```diff
@@ -204,24 +205,36 @@ def test_pbs_split_2_job_nb_commands(self):
...
+    def setUp(self):
+        self._home_backup = os.environ['HOME']
+        os.environ['HOME'] = tempfile.mkdtemp()
 
-    job_generator = job_generator_factory(queue, commands, cluster_name="helios")
-    assert_true(isinstance(job_generator, HeliosJobGenerator))
+        self.rap_filename = os.path.join(os.environ['HOME'], ".default_rap")
+        if os.path.isfile(self.rap_filename):
+            raise Exception("Test fail: {} should not be there.".format(self.rap_filename))
+        else:
```
MarcCote: No need for an `else` here. Just de-indent the next couples lines.

MarcCote:
Except for some minor comments, LGTM.

smartdispatch/job_generator.py
mgermain: Yeah you're right.
At first. I felt that the followint look weird for some reason. I'll change it
```
def _add_cluster_specific_rules():
    pass
```

smartdispatch/job_generator.py
mgermain: Ok

smartdispatch/tests/test_job_generator.py
mgermain: Yes, it is still called. We were using 2 different testing frameworks that are almost 100% compatible with each other.

smartdispatch/tests/test_job_generator.py
mgermain: Good question. I'll change it.

smartdispatch/tests/test_job_generator.py
mgermain: No need indeed but I think it's much clearer to put it in the else.
It put the emphasis on the fact that those lines will only be run when the condition is false.

MarcCote: ok

smartdispatch/tests/test_job_generator.py
MarcCote: ok, good to know.

MarcCote:
Thanks.

PR #126, opened 2016-04-19
More robust file locking

mgermain:


smartdispatch/filelock.py
```diff
@@ -56,15 +67,19 @@ def open_with_dirlock(*args, **kwargs):
...
         except OSError:
-            logging.info("Can't immediately write-lock the file ({0}), retrying in {1} sec. ...".format(filename, TIME_BETWEEN_ATTEMPTS))
+            logging.info("Can't immediately write-lock the file ({0}), retrying in {1} sec.".format(filename, TIME_BETWEEN_ATTEMPTS))
             time.sleep(TIME_BETWEEN_ATTEMPTS)
             no_attempt += 1
 
+    if no_attempt == MAX_ATTEMPTS:
+        raise IOError("Failed to lock {0} {1} times.".format(f.name, MAX_ATTEMPTS))
+
+    yield f
```
MarcCote: I just realized, if there is an exception that is raised when yielding the lock will stay there forever. We should use a try-finally to always remove the lock. Something like
```python

try:
    with open(*args, **kwargs) as f:
        yield f
finally:
    os.rmdir(lockfile)
```

smartdispatch/filelock.py
```diff
@@ -56,15 +67,19 @@ def open_with_dirlock(*args, **kwargs):
         try:
             os.mkdir(lockfile)  # Atomic operation
             f = open(*args, **kwargs)
```
MarcCote: We shoud move this below. See my comment at line 79.

smartdispatch/filelock.py
```diff
@@ -36,8 +36,19 @@ def open_with_flock(*args, **kwargs):
...
         fcntl.lockf(f, fcntl.LOCK_EX | fcntl.LOCK_NB)
     except IOError:
-        logging.info("Can't immediately write-lock the file ({0}), waiting ...".format(f.name))
-        fcntl.lockf(f, fcntl.LOCK_EX)
+        no_attempt = 0
+        while no_attempt < MAX_ATTEMPTS:
+            try:
+                logging.info("Can't immediately write-lock the file ({0}), waiting.".format(f.name))
+                start_time = time.time()
+                fcntl.lockf(f, fcntl.LOCK_EX)
```
MarcCote: What is the error raised when the `f` doesn't exist? If it is a `IOError` we might be in trouble-ish, it will loop until `MAX_ATTEMPTS`.

MarcCote:
I like the idea but we should test this correctly before merging this fallback mechanism into master.

smartdispatch/filelock.py
mgermain: You are right. I should be more specific.

smartdispatch/filelock.py
mgermain: Right

MarcCote:
Great.

PR #127, opened 2016-04-19
Small bug fix

mgermain:
First, there was 2 missing blank lines in the README.

Second, we were doing this:
```
get_available_queues(cluster_name=utils.detect_cluster()):
```

Which was calling `qstat` every time this function was imported.

And `smartdispatch/__init__.py` was importing it `from smartdispatch.smartdispatch import *`.

So `qstat` was called way too often for no reasons.

### DIFF
--- a/smartdispatch/smartdispatch.py
+++ b/smartdispatch/smartdispatch.py
@@ -105,7 +105,7 @@ def replace_uid_tag(commands):
     return [command.replace("{UID}", utils.generate_uid_from_string(command)) for command in commands]
 
 
-def get_available_queues(cluster_name=utils.detect_cluster()):
+def get_available_queues(cluster_name):
     """ Fetches all available queues on the current cluster """
     if cluster_name is None:
         return {}
--- a/README.md
+++ b/README.md
@@ -44,7 +44,8 @@ python my_script.py 2 80 tanh 0.1
 ```
 
 Another possiblility is to generate argument from a range.
-``smart-dispatch -q qtest@mp2 launch python my_script.py [1:4]``
+
+`smart-dispatch -q qtest@mp2 launch python my_script.py [1:4]`
 
 Will generate:
 ```
@@ -54,7 +55,8 @@ python my_script.py 3
 ```
 
 You can also add a step size to the range as the 3rd argument.
-``smart-dispatch -q qtest@mp2 launch python my_script.py [1:10:2]``
+
+`smart-dispatch -q qtest@mp2 launch python my_script.py [1:10:2]`
 
 Will generate:
 ```


smartdispatch/smartdispatch.py
```diff
@@ -105,7 +105,7 @@ def replace_uid_tag(commands):
     return [command.replace("{UID}", utils.generate_uid_from_string(command)) for command in commands]
 
 
-def get_available_queues(cluster_name=utils.detect_cluster()):
+def get_available_queues(cluster_name):
```
MarcCote: I think it would be better to have:
```python
def get_available_queues(cluster_name=None):
    if cluster_name is None:
        cluster_name = utils.detect_cluster()
...
```

MarcCote:
Maybe I'm missing something but in Python a module get initialized only once even if it is called several times. See http://stackoverflow.com/a/296062

smartdispatch/smartdispatch.py
mgermain: Maybe but we would have to change the tests and the detect_cluster() function.
I'm under the impression that having a default for this function is probably not a good idea.

mgermain:
@MarcCote you might be right. I was trying this error that happen sometime. 
```
pbs_iff: cannot connect to mp2:15001 - timeout, errno=99 (Cannot assign requested address)
No Permission.
qstat: cannot connect to server mp2 (errno=15007) Unauthorized Request
Traceback (most recent call last):
  File "/home/germainm/smartdispatch/smartdispatch/workers/base_worker.py", line 10, in <module>
    from smartdispatch import utils
  File "/home/germainm/smartdispatch/smartdispatch/__init__.py", line 2, in <module>
    from smartdispatch.smartdispatch import *
  File "/home/germainm/smartdispatch/smartdispatch/smartdispatch.py", line 108, in <module>
    def get_available_queues(cluster_name=utils.detect_cluster()):
  File "/home/germainm/smartdispatch/smartdispatch/utils.py", line 99, in detect_cluster
    server_name = output.split('\n')[2].split(' ')[0]
IndexError: list index out of range
```

It might not be the right way but the `def get_available_queues(cluster_name=utils.detect_cluster()):` does not feel clean. Is there any standard/good practice with default value of named parameter and function ?

MarcCote:
I agree with you `get_available_queues(cluster_name=utils.detect_cluster())` was a bad idea. Regarding the standard usually what I see is either have a standard type or `None` as default values. When it is `None`, a check is performed inside the function to compute the default value. This is useful when you have to call another function to compute such value, or even when you expect the parameter to be a list that you plan on modifying in-place.

smartdispatch/smartdispatch.py
MarcCote: Why do you have to change the tests and `detect_cluster()`? But you might be right about simply not having a default for this function.

MarcCote:
Alright, the discussion about default value and initialization of module are interesting but as @mgermain has pointed out we don't really need default value for the function `get_available_queues` anyway. So, I'm merging this one. Thanks.

smartdispatch/smartdispatch.py
mgermain: We test that None return an empty list.

MarcCote: It would have still returned an empty list because of `cluster_name = utils.detect_cluster()` returns `None`... oh right only when not on a cluster. You right :P

Issue #129, opened 2016-05-11
extra jobs for space


SinaHonari:
The `-f` feature launch extra job when there is a blank line.

```
python create_procs.py --L2_coef=1e-08 --L2_coef_ful=1e-10 --file_suffix=test1

python create_procs.py --L2_coef=1e-08 --L2_coef_ful=1e-08 --file_suffix=test2
```

This launches 3 jobs and it should launch 2.

PR #134, opened 2016-07-29
Bug fix for issue #129

Dutil:


smartdispatch/smartdispatch.py
```diff
@@ -59,7 +59,8 @@ def get_commands_from_file(fileobj):
     commands : list of str
         commands read from the file
     '''
-    return fileobj.read().strip().split('\n')
+
+    return filter(None, fileobj.read().strip().split('\n'))
```
MarcCote: The `filter` function in Python 3 returns a generator instead of a list. I know that right now we are not supporting Python3 but we could avoid a potential error by using a list comprehension here.

I think the following is equivalent: 
```python
return [line.strip() for line in fileobj if len(line.strip()) > 0]
```

mgermain:
Perfect! Add some unit tests and I'll merge it.

PR #135, opened 2016-08-19
print job and node in log

### LINKED ISSUE #133: Add more info to the commands log.
The line to modify are the 2 prints there :
https://github.com/SMART-Lab/smartdispatch/blob/master/smartdispatch/workers/base_worker.py#L50-L52

``## SMART-DISPATCH - Started on: 2016-07-12 16:28:23 ##``
Should look like:
``## SMART-DISPATCH - Started on: 2016-07-12 16:28:23 - In job: {jobid} - On node: {nodeid} ##``

slefrancois:
fixes #133

### DIFF
--- a/smartdispatch/workers/base_worker.py
+++ b/smartdispatch/workers/base_worker.py
@@ -46,25 +46,14 @@ def main():
         stderr_filename = os.path.join(args.logs_dir, uid + ".err")
 
         # Get job and node ID
-        try:
-            job = os.environ['PBS_JOBID']
-        except KeyError:
-            job = 'undefined'
-
-        try:
-            with open(os.environ['PBS_NODEFILE']) as nodefile:
-                node = ""
-                for line in nodefile:
-                    node += line + ','
-                node = str.rstrip(node, ' ,\n')
-        except KeyError:
-            node = 'undefined'
+        job_id = os.environ.get('PBS_JOBID', 'undefined')
+        node_name = os.environ.get('HOSTNAME', 'undefined')
 
         with open(stdout_filename, 'a') as stdout_file:
             with open(stderr_filename, 'a') as stderr_file:
-                log_datetime = t.strftime("## SMART-DISPATCH - Started on: %Y-%m-%d %H:%M:%S - In job: {job} - On nodes: {node} ##\n".format(job=job, node=node))
+                log_datetime = t.strftime("## SMART-DISPATCH - Started on: %Y-%m-%d %H:%M:%S - In job: {job_id} - On nodes: {node_name} ##\n".format(job_id=job_id, node_name=node_name))
                 if stdout_file.tell() > 0:  # Not the first line in the log file.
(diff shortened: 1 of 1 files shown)


MarcCote:
We should use a default value if either `PBS_JOBID` or `PBS_NODEFILE` doesn't exist.

smartdispatch/workers/base_worker.py
```diff
@@ -45,11 +45,26 @@ def main():
         stdout_filename = os.path.join(args.logs_dir, uid + ".out")
         stderr_filename = os.path.join(args.logs_dir, uid + ".err")
 
+        # Get job and node ID
+        try:
+            job = os.environ['PBS_JOBID']
```
MarcCote: Instead of using a try catch, we can use the `get` method.
`job = os.environ.get('PBS_JOBID', 'undefined')`

smartdispatch/workers/base_worker.py
```diff
@@ -45,11 +45,26 @@ def main():
...
+        except KeyError:
+            job = 'undefined'
+
+        try:
+            with open(os.environ['PBS_NODEFILE']) as nodefile:
```
MarcCote: Nice use of the try catch here :+1: .

slefrancois: Which should we pick, try/catch or get? get avoids having to manage exceptions.

MarcCote: Here we can't really use the `get` method as above. What would be the default value for the file to open? I think the try/catch is well suited for this case.

Actually, we might want to catch a `IOError` as well, in case `os.environ['PBS_NODEFILE']` is not a file.

slefrancois: Good point, I'll clean it up next week.

mgermain: There is a better way to do this.
```node_name = os.environ.get('HOSTNAME', 'undefined')```

File: smartdispatch/workers/base_worker.py [hunk 3]
```diff
@@ -45,11 +45,26 @@ def main():
         stdout_filename = os.path.join(args.logs_dir, uid + ".out")
         stderr_filename = os.path.join(args.logs_dir, uid + ".err")
 
+        # Get job and node ID
+        try:
+            job = os.environ['PBS_JOBID']
+        except KeyError:
+            job = 'undefined'
```
mgermain: job_id

smartdispatch/workers/base_worker.py
slefrancois: It does work, but how about when multiple nodes are assigned to a job? The nodefile will have multiple entries, but $HOSTNAME I'm not sure.

slefrancois:
I used the HOSTNAME env variable to get the node for simplicity. If the code is extended to support multi-node jobs, it may be more reliable to use the PBS_NODEFILE instead:
https://gist.github.com/slefrancois/e30801f267bfabb492e202f383a59fff

MarcCote:
LGTM

slefrancois:
Moving forward, there is an issue #137 with some clusters like Helios where the PBS job ID from Torque and MOAB don't match.

PR #136, opened 2016-08-19
check that requested cores >= 1 and < node core total

### LINKED ISSUE #92: Validate # of core per command
Tell the user that his job can't be launched on this queue if one command requires more cores than 1 node can provide.

Let say one node of a cluster has 24 cores, if a user set --coresPerCommand 30 it should alert him and not launch the jos.

We should also verify that it cannot be set lower than 1.

slefrancois:
Fixes #92

### DIFF
--- a/scripts/smart-dispatch
+++ b/scripts/smart-dispatch
@@ -113,6 +113,12 @@ def main():
     # TODO: use args.memPerNode instead of args.memPerNode
     queue = Queue(args.queueName, CLUSTER_NAME, args.walltime, args.coresPerNode, args.gpusPerNode, float('inf'), args.modules)
 
+    # Check that requested core number does not exceed node total
+    if args.coresPerCommand > queue.nb_cores_per_node:
+        sys.stderr.write("smart-dispatch: error: coresPerCommand exceeds nodes total: asked {req_cores} cores, nodes have {node_cores}\n"
+                         .format(req_cores=args.coresPerCommand, node_cores=queue.nb_cores_per_node))
+        sys.exit(2)
+
     command_params = {'nb_cores_per_command': args.coresPerCommand,
                       'nb_gpus_per_command': args.gpusPerCommand,
                       'mem_per_command': None  # args.memPerCommand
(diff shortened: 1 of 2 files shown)


scripts/smart-dispatch
```diff
@@ -113,6 +113,12 @@ def main():
     # TODO: use args.memPerNode instead of args.memPerNode
     queue = Queue(args.queueName, CLUSTER_NAME, args.walltime, args.coresPerNode, args.gpusPerNode, float('inf'), args.modules)
 
+    # Check that requested core number does not exceed node total
+    if args.coresPerCommand > queue.nb_cores_per_node:
+        sys.stderr.write("smart-dispatch: error: coresPerCommand exceeds nodes total: asked {req_cores} cores, nodes have {node_cores}\n"
+                         .format(req_cores=args.coresPerCommand, node_cores=queue.nb_cores_per_node))
+        sys.exit(2)
```
MarcCote: Is there any particular reason for `sys.exit(2)` instead of `sys.exit(1)` ?

scripts/smart-dispatch
```diff
@@ -175,6 +181,8 @@ def parse_arguments():
             parser.error("You need to specify a command to launch.")
         if args.queueName not in AVAILABLE_QUEUES and ((args.coresPerNode is None and args.gpusPerNode is None) or args.walltime is None):
             parser.error("Unknown queue, --coresPerNode/--gpusPerNode and --walltime must be set.")
+        if args.coresPerCommand < 1:
+            parser.error("coresPerNode must be at least 1")
```
MarcCote: Good catch.

MarcCote:
Can you make unit tests to check your fix?

scripts/smart-dispatch
slefrancois: 2 is usually the code for a command line error, it's what parser.error() returns.

MarcCote: Cool, didn't know that.

mgermain:
Yup just add some tests in `test_smart_dispatch.py` and I'm good with merging this.

MarcCote:
LGTM

Issue #11, opened 2014-02-20
Unique name for each batch of jobs


MarcCote:
Right now, if we launch multiple batch of jobs (batch => call to smartDispatch.py) the same Name will be reused.

| Job id | Name | User | Time Use | S | Queue |
| --- | --- | --- | --- | --- | --- |
| 24834.mp2 | jobCommands_0.sh | stjeans1 | 50:58:21 | R | qwork4 |
| 24835.mp2 | jobCommands_1.sh | stjeans1 | 64:02:49 | R | qwork4 |
| 24836.mp2 | jobCommands_2.sh | stjeans1 | 43:58:09 | R | qwork4 |
| 24837.mp2 | jobCommands_0.sh | stjeans1 | 51:58:06 | R | qwork4 |
| 24838.mp2 | jobCommands_1.sh | stjeans1 | 23:22:10 | R | qwork4 |
| 24839.mp2 | jobCommands_2.sh | stjeans1 | 63:27:05 | R | qwork4 |

mgermain:
Putting the commandUID won't work because there can by as many command in a job as the user want. So in the end I think it should be {batchUID}_# instead of jobCommands_#.sh

We should probably use the PBS option jobNAME and juste name the files job_1.sh ... or name them the same ?

MarcCote:
I would prefer naming them the same using {batchUID}# as you proposed.

mgermain:
You mean having the file AND the jobname set to {natchUID} ?

MarcCote:
Thinking over, let's go with what you proposed. That is, using PBS option -jobname (needs to check what is the actual name here) to set {batchUID}# as the name of the PBS job. Regarding the files contained in `{batchUID}/commands/` we can let them to `job_#.sh`.

negar-rostamzadeh:
Since batchUID contains the name of the user and the name of the script, with the suggested naming system, we will end up with having long names. Then what will happen is that: when we call qstat, we are usually just seeing the end of the file name. Then we will running to the same issue when we are checking the jobs using qstat (when we have multiple batches of jobs):

138498.gpu-srv1            ...menegartestsh_0 negar           00:00:00 C test           
138499.gpu-srv1            ...menegartestsh_1 negar           00:00:00 C test           
138500.gpu-srv1            ...menegartestsh_0 negar           00:00:00 C test  
-----------------------------
So do you propose a better job naming system that be both representative of the job and also short enough to prevent confusion in qstat output?

PR #144, opened 2016-11-11
unique names for each job in multiple batches

negar-rostamzadeh:
@mgermain 
regarding to the #11

### DIFF
--- a/scripts/smart-dispatch
+++ b/scripts/smart-dispatch
@@ -128,6 +128,10 @@ def main():
                       }
 
     job_generator = job_generator_factory(queue, commands, command_params, CLUSTER_NAME, path_job)
+    for pbs in job_generator.pbs_list:
+        pbs.add_options(options='-N ' + jobname + '_' + str(i))
+        i += 1
+
     if args.pbsFlags is not None:
         job_generator.add_pbs_flags(args.pbsFlags.split(' '))
     pbs_filenames = job_generator.write_pbs_files(path_job_commands)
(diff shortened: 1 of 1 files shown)


negar-rostamzadeh:
The test is failed as I changed the argument format in smart_dispatch? do you suggest to change the launcher in a different file that can be passed to the smart_dispatch script?

scripts/smart-dispatch
```diff
@@ -136,9 +136,12 @@ def main():
     print "## {nb_commands} command(s) will be executed in {nb_jobs} job(s) ##".format(nb_commands=nb_commands, nb_jobs=len(pbs_filenames))
     print "Batch UID:\n{batch_uid}".format(batch_uid=jobname)
     if not args.doNotLaunch:
+        in_batch_job_counter = -1
         jobs_id = []
         for pbs_filename in pbs_filenames:
-            qsub_output = check_output('{launcher} {pbs_filename}'.format(launcher=LAUNCHER if args.launcher is None else args.launcher, pbs_filename=pbs_filename), shell=True)
+            in_batch_job_counter + = 1
```
MarcCote: The test is failing because this is not a valid Python syntax. See https://travis-ci.org/SMART-Lab/smartdispatch/builds/175228382#L230

MarcCote:
Of course, the unit tests should reflect the changes that were made.

scripts/smart-dispatch
```diff
@@ -128,6 +128,10 @@ def main():
                       }
 
     job_generator = job_generator_factory(queue, commands, command_params, CLUSTER_NAME, path_job)
+    for pbs in job_generator.pbs_list:
+        pbs.add_options(options='-N ' + jobname + '_' + str(i))
+        i += 1
```
mgermain: You forgot to initialize the `i`

scripts/smart-dispatch
```diff
@@ -128,6 +128,10 @@ def main():
                       }
 
     job_generator = job_generator_factory(queue, commands, command_params, CLUSTER_NAME, path_job)
+    for pbs in job_generator.pbs_list:
+        pbs.add_options(options='-N ' + jobname + '_' + str(i))
```
mgermain: This will add `options` to the PBS file this is not what you want.
`pbs.add_options(N=jobname + '_' + str(i))`

mgermain:
Don't forget the unit tests.

scripts/smart-dispatch
mgermain: If you take a look at https://github.com/SMART-Lab/smartdispatch/blob/master/smartdispatch/pbs.py#L58
you will see this will fail quite often. You should add a function in `utils.py` that takes a jobname and a id and return a string of length 64. Also, add a small test for that function.

negar-rostamzadeh:
sure, I will change it in the utils then. But I don't still understand why
the name shouldn't be added through the options? Then it will be useful for
the later changes to add the functionality of adding a manual name through
options instead of just applying the default names.

PR #145, opened 2016-11-18
Autoresume feature

ddtm:
Implements #138 

A user can now add `--autoresume` to automatically requeue her jobs if the running time exceeds maximum walltime allowed on the cluster.


smartdispatch/tests/test_pbs.py
```diff
@@ -100,11 +102,13 @@ def test_str(self):
...
+# Prolog #
+
 # Commands #
-{command1} &
-{command2} &
+{command1}
+{command2}
 
-wait""".format(command1=commands[0], command2=commands[1])
+# Epilog #""".format(command1=commands[0], command2=commands[1])
```
MarcCote: We should add a prolog and an epilog when testing a more complex PBS file (like we do with the commands).

smartdispatch/workers/base_worker.py
```diff
@@ -35,8 +37,24 @@ def main():
...
+            return
+        else:
+            sigterm_handler.triggered = True
+        if sigterm_handler.command is not None:
+            command_manager.set_running_command_as_pending(sigterm_handler.command)
+        sys.exit(0)
+
+    sigterm_handler.triggered = False
+    sigterm_handler.command = None
+    signal.signal(signal.SIGTERM, sigterm_handler)
```
MarcCote: Is it possible somehow the worker receives a SIGTERM even though we didn't use `--autoresume` ? What kind of signal is the scheduler sending by default when max_walltime has been reached?

If it is the case that the scheduler sends a SIGTERM, that would mean all unfinished commands are being sent back to the pending list. Are we guaranteed there would be enough time to do it?

Also, without the `--autoresume`, I would rather have the commands sent to a `unfinished_commands.txt` file instead of sent back into pending. Not all commands are resumable.

smartdispatch/workers/base_worker.py
```diff
@@ -35,8 +37,24 @@ def main():
 
     command_manager = CommandManager(args.commands_filename)
 
+    # Handle TERM signal gracefully by sending running commands to back to
+    # the list of pending commands.
+    def sigterm_handler(signal, frame):
+        if sigterm_handler.triggered:
```
MarcCote: Multiple `SIGTERM` signals can be received, how is that?

scripts/smart-dispatch
```diff
@@ -28,6 +28,21 @@ CLUSTER_NAME = utils.detect_cluster()
 AVAILABLE_QUEUES = get_available_queues(CLUSTER_NAME)
 LAUNCHER = utils.get_launcher(CLUSTER_NAME)
 
+AUTORESUME_EPILOG = """\
+NEED_TO_RESUME=false
+for WORKER_PID in $WORKER_PIDS; do
+    wait "$WORKER_PID"
+    RETURN_CODE=$?
+    if [ $RETURN_CODE -eq 124 ]; then
```
MarcCote: What does the return code 124 means? I'm guessing it has something to do with the timeout command? Maybe add a comment mentioning it.

scripts/smart-dispatch
```diff
@@ -107,10 +122,20 @@ def main():
 
     # Generating all the worker commands
     worker_script = pjoin(os.path.dirname(smartdispatch.__file__), 'workers', 'base_worker.py')
-    COMMAND_STRING = 'cd "{cwd}"; python2 {worker_script} "{commands_file}" "{log_folder}" '\
+
+    worker_call_prefix = ''
+    worker_call_suffix = ''
+    if args.autoresume:
+        worker_call_prefix = 'timeout -s TERM $(($PBS_WALLTIME - 60)) '
+        worker_call_suffix = ' WORKER_PIDS+=" $!"'
```
MarcCote: Neat!

scripts/smart-dispatch
```diff
@@ -28,6 +28,21 @@ CLUSTER_NAME = utils.detect_cluster()
...
+for WORKER_PID in $WORKER_PIDS; do
+    wait "$WORKER_PID"
+    RETURN_CODE=$?
+    if [ $RETURN_CODE -eq 124 ]; then
+        NEED_TO_RESUME=true
+    fi
+done
+if [ "$NEED_TO_RESUME" = true ]; then
+    echo "Autoresuming using: {launcher} $PBS_FILENAME"
+    {launcher} $PBS_FILENAME
```
MarcCote: Neat!

smartdispatch/workers/base_worker.py
ddtm: List of unfinished jobs is a good idea (we discussed that with @mgermain) but it is orthogonal to this PR.

60s (default) should be enough to update the lists.

MarcCote: Assuming SIGTERM is also sent by the scheduler, you won't have 60s unless you provided `--autoresume` to smart-dispatch, right?

MarcCote:
Is there any reason we didn't try setting an alarm when launching the PBS files with msub?
https://computing.llnl.gov/tutorials/moab/#TimeExpired

I'm guessing the answer is that the option is not supported by qsub?

Anyhow, if the worker are indeed receiving two SIGTERM signals, maybe we should send a SIGALRM instead?

smartdispatch/workers/base_worker.py
mgermain: According to [this](https://computing.llnl.gov/LCdocs/moab/index.jsp?show=s5.3.1), the default grace time is 60 seconds. I could not find the info for helios in particular but we can request a specific grace time (we need to check if it'S the same with qsub).

We could always explicitly ask for a 1 minute grace signal?

mgermain:
@MarcCote  

 -notify
          Available for qsub,  qrsh  (with  command)  and  qalter
          only.

          This flag, when set causes  Sun  Grid  Engine  to  send
          "warning" signals to a running job prior to sending the
          signals themselves. If a SIGSTOP is  pending,  the  job
          will  receive a SIGUSR1 several seconds before the SIG-
          STOP. If a SIGKILL is pending, the job will  receive  a
          SIGUSR2  several  seconds  before  the  SIGKILL.   .....

mgermain:
@ddtm @MarcCote 

Did some quick testing.
`smart-dispatch -q gpu_1 --walltime=0:03:00 --pbsFlags="-lsignal=14@120" launch python -u test_signal.py`
```
Fri Nov 18 18:01:33 2016 - Started
Fri Nov 18 18:02:28 2016 - 14 SIGALRM
Fri Nov 18 18:02:28 2016 - 14 SIGALRM
Fri Nov 18 18:02:28 2016 - 14 SIGALRM
Fri Nov 18 18:02:28 2016 - 14 SIGALRM
Fri Nov 18 18:02:28 2016 - 14 SIGALRM
Fri Nov 18 18:02:28 2016 - 14 SIGALRM
Fri Nov 18 18:02:28 2016 - 15 SIGTERM
```

There is something wrong happening, I'll investigate more.
First the sigalarm whould ne be there 6 times. (It's always 6)
Second the sitterm is arriving 1 minute early when I use sigterm.

This was on helios.

smartdispatch/workers/base_worker.py
MarcCote: Sure. But still, there is a change in the behavior of smart-dispatch. Unresumable commands (default) shouldn't be put back into the pending list.

mgermain: Why? It's a much better behavior than we have now. We already put it back in pending, we just put it at a later time.
As I discussed with @ddtm and you suggested it would be best to put them back in unfinished but it's a separate issue that should be done in a different PR.

MarcCote: No doubt about the unfinished list being a good idea and I also agree it should be done in a different PR.

My point is right now, looking at the remaining commands in running.txt is the only way I've got to know which commands are unfinished. I don't want them to go back into the pending list because they are **not** resumable.

@mgermain when you say "We already put it back in pending, we just put it at a later time." that only happens if you use the resume subcommand of smart-dispatch.

ddtm: When one uses `--autoresume`, it's her responsibility to make sure that commands are resumable.

MarcCote: I think my first post wasn't clear. Let me try to clarify it:

Is it possible somehow the worker receives a SIGTERM even though we **didn't** use --autoresume ?

From what @mgermain showed (see https://github.com/SMART-Lab/smartdispatch/pull/145#issuecomment-261666784), it seems so. The workers receive a SIGTERM when their execution walltime exceeds the max_walltime. 

That would mean that **every** command will be put back into the pending list whether or not `--autoresume` has been used.

MarcCote:
@mgermain Is it possible that on helios the period of grace is actually two minutes instead of 60 seconds (as you or @ddtm mentioned before)? 

Can you try: `smart-dispatch -q gpu_1 --walltime=0:04:00 --pbsFlags="-lsignal=14@90" launch python -u test_signal.py`

smartdispatch/workers/base_worker.py
```diff
@@ -35,8 +37,24 @@ def main():
...
+            return
+        else:
+            sigterm_handler.triggered = True
+        if sigterm_handler.command is not None:
+            command_manager.set_running_command_as_pending(sigterm_handler.command)
+        sys.exit(0)
+
+    sigterm_handler.triggered = False
+    sigterm_handler.command = None
+    signal.signal(signal.SIGTERM, sigterm_handler)
```
MarcCote: Let's add a `--auto-resume` to this script too and set the signal handler **only if** `--auto-resume` is true. Of course, this new option should be used accordingly in the script `smart-dispatch`.

smartdispatch/workers/base_worker.py
```diff
@@ -37,24 +38,28 @@ def main():
...
+                return
+            else:
+                sigterm_handler.triggered = True
+            if sigterm_handler.proc is not None:
+                sigterm_handler.proc.wait()
+            if sigterm_handler.command is not None:
+                command_manager.set_running_command_as_pending(sigterm_handler.command)
+            sys.exit(0)
+
+        sigterm_handler.triggered = False
```
mgermain: Set sigterm_handler.proc to None here.

smartdispatch/workers/base_worker.py
```diff
@@ -17,6 +17,7 @@ def parse_arguments():
     parser = argparse.ArgumentParser()
     parser.add_argument('commands_filename', type=str, help='File containing all commands to execute.')
     parser.add_argument('logs_dir', type=str, help="Folder where to put commands' stdout and stderr.")
+    parser.add_argument('-r', '--assume_resumable', action='store_true', help="Assume that commands are resumable and put them into the pending list on worker termination.")
```
mgermain: Should use camelcase.

smartdispatch/workers/base_worker.py
```diff
@@ -35,8 +38,28 @@ def main():
...
+    if args.assume_resumable:
+        # Handle TERM signal gracefully by sending running commands back to
+        # the list of pending commands.
+        def sigterm_handler(signal, frame):
+            if sigterm_handler.triggered:
+                return
+            else:
+                sigterm_handler.triggered = True
+            if sigterm_handler.proc is not None:
+                sigterm_handler.proc.wait()
```
mgermain: You should also take a look at the return code and put it to failed it it fail while handling the signal.

MarcCote:
I have some experiments to do, so I'll try this awesome feature right away. :)

scripts/smart-dispatch
```diff
@@ -28,6 +28,27 @@ CLUSTER_NAME = utils.detect_cluster()
...
+for WORKER_PID in $WORKER_PIDS; do
+    wait "$WORKER_PID"
+    RETURN_CODE=$?
+    if [ $RETURN_CODE -eq {timeout_exit_code} ]; then
+        NEED_TO_RESUME=true
+    fi
+done
+if [ "$NEED_TO_RESUME" = true ]; then
+    echo "Autoresuming using: {{launcher}} $PBS_FILENAME"
+    {{launcher}} $PBS_FILENAME
```
MarcCote: After testing it a couple of times, I can say it the jobs are relaunched correctly. However, the `jobs_ids.txt` in the experiment folder is not being updated.  Can you think of a way to update it accordingly? (See lines related to https://github.com/SMART-Lab/smartdispatch/pull/145/files#diff-8a01797a6fa72cb758a419e28138d1f0R183)

scripts/sd-launch-pbs
```diff
@@ -0,0 +1,36 @@
...
+
+    args = parse_arguments()
+
+    launch_jobs(LAUNCHER if args.launcher is None else args.launcher, [args.pbs], CLUSTER_NAME, args.pathJob)
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-L', '--launcher', choices=['qsub', 'msub'], required=False, help='Which launcher to use. Default: qsub')
+    parser.add_argument('-p', '--pbs', required=True, help='PBS filename to launch.')
```
MarcCote: Since this a mandatory argument, we shouldn't use `--`.

That would give:
`parser.add_argument(pbs, help='PBS filename to launch.')`

Make sure this script is called accordingly after the change.

scripts/sd-launch-pbs
```diff
@@ -0,0 +1,36 @@
...
+    parser.add_argument('-j', '--pathJob', required=True, help='Path to the job folder.')
```
MarcCote: Same comment as the previous argument.

smartdispatch/workers/base_worker.py
```diff
@@ -35,8 +38,37 @@ def main():
...
+            if sigterm_handler.triggered:
+                return
+            else:
+                sigterm_handler.triggered = True
+            error_code = 0
+            if sigterm_handler.proc is not None:
+                error_code = sigterm_handler.proc.wait()
+            if sigterm_handler.command is not None:
+                if error_code == 0:  # The command was terminated successfully.
+                    command_manager.set_running_command_as_pending(sigterm_handler.command)
```
MarcCote: Why are the commands, that have terminated successfully, sent to pending? Is it because when they receive a sigterm signal, they have an error code of 0?

MarcCote:
I'm using since last night and my experiments have resumed successfully many times. @mgermain all good on my side.

MarcCote:
@mgermain anything you want to add? Otherwise you can go ahead and merge it. Thanks again @ddtm

mgermain:
Everything is fine now. But for cleanliness we should add the `-l depends=<current_jobid>` to the relaunched jobs.

MarcCote:
Also, a rebase seems to be needed.

PR #148, opened 2016-11-25
Defaultn

negar-rostamzadeh:
jobname is not a good name and should be changed to Batch_UID to be more representative


mgermain:
@negar-rostamzadeh Be careful, you committed the whole SMART_DISPATCH_LOGS folder. 12 log files that should be removed.

smartdispatch/utils.py
```diff
@@ -17,6 +17,21 @@ def print_boxed(string):
...
 
+def jobname_generator(jobname, job_id):
+    """check if the length of the jobname is less than 64"""
+    
+    """args:
+	    jobname: initial jobname
+	    job_id: id of the job in the current batch
+	    
+      returns: 
+	    croped_string: the croped version of the string"""
```
mgermain: Please use the proper numpy standard for docstring as used in the rest of the program described [here](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt#docstring-standard)

Something like this:
```
"""Crop the jobname to a maximum of 64 characters.

Parameters
----------
jobname : str
    Initial jobname.
job_id: str
    ID of the job in the current batch.

Returns
-------
str
    The cropped version of the string.	
"""
```

scripts/smart-dispatch
```diff
@@ -128,6 +128,14 @@ def main():
                       }
 
     job_generator = job_generator_factory(queue, commands, command_params, CLUSTER_NAME, path_job)
+    ctr = 0 
+    # generating default names per each jobs in each batch
+    for pbs in job_generator.pbs_list:
+	    proper_size_name = utils.jobname_generator(jobname, str(ctr)) 
+	    pbs.add_options(N = proper_size_name)
```
mgermain: Please follow PEP8 everywhere, in this case follow [this one](https://www.python.org/dev/peps/pep-0008/#other-recommendations). Nos spaces around named parameters.

smartdispatch/utils.py
```diff
@@ -17,6 +17,21 @@ def print_boxed(string):
...
+    if len(jobname) + len(job_id) > 64:
+	extra_length =  len(jobname) + len(job_id) - 64
```
mgermain: Bad indentation, this should not even run...
Also you mix tab and spaces.

smartdispatch/utils.py
```diff
@@ -17,6 +17,21 @@ def print_boxed(string):
...
+def jobname_generator(jobname, job_id):
+    """check if the length of the jobname is less than 64"""
+    
+    """args:
+	    jobname: initial jobname
+	    job_id: id of the job in the current batch
+	    
+      returns: 
+	    croped_string: the croped version of the string"""
+    if len(jobname) + len(job_id) > 64:
```
mgermain: This might still break, you are not considering the `_`.

smartdispatch/utils.py
mgermain: Actually, you don't need this line look at the next comment.

mgermain:
@negar-rostamzadeh You still forgot to put the tests.

PR #159, opened 2017-02-07
Don't export the user's environment to the job

### LINKED ISSUE #157: Don't use -V
Currently smart-dispatch exports the current environment to the job using `-V`. This is bad practice, see e.g. McGill's HPC guidelines:

> We no longer recommend adding corresponding "module load ..." lines to your .bashrc file, or using "#PBS -V". A job is most reliably re-run if it is self-contained and has all information needed, instead of relying on the environment.

A more concrete example: Helios's Lua-based module system is broken, conflicting with any present Torch installation. The use of `-V` causes all module loading to fail whenever Torch was active in the launching environment.

bartvm:
Fixes #157.

Currently the user's entire environment is exported. This is bad
practice because it makes jobs hard to reproduce and debug, and in some
cases it can even cause errors. The only variable that needs to be
exported is `PBS_FILENAME`, which is used by the auto-resume script.

### DIFF
--- a/smartdispatch/pbs.py
+++ b/smartdispatch/pbs.py
@@ -36,7 +36,7 @@ class PBS(object):
         self.add_options(q=queue_name)
 
         # Declares that all environment variables in the qsub command's environment are to be exported to the batch job.
-        self.add_options(V="")
+        self.add_options(v="PBS_FILENAME")
 
     def add_options(self, **options):
         """ Adds options to this PBS file.


bartvm:
@mgermain Test away.

bartvm:
@mgermain Ping.

mgermain:
@bartvm working on one cluster, not working on the other. Haven't had the time to investigate why yet.

bartvm:
Do you get an error? Or does it just not restart?

mgermain:
Without the -V on Mammouth some part of the environment is not loaded.

bartvm:
That's the entire point of the flag, no? But smart-dispatch shouldn't rely on the environment as set outside of the job. What environment variable does it try to access?

mgermain:
No, the point of the flag is to not copy the **current** environment.
Your system still has to load the environment of the user in the job. It does on Helios and it does it weirdly and partially on mp2.

It does not try to access anything, I just saw that some modules are missing when -V is used. I have to investigate why.

bartvm:
That sounds weird... `-V` shouldn't affect at all how the job's environment is set up beyond not importing environment variables from the launch environment. It's possible that Mammouth doesn't load `.bashrc` at the beginning of a job. But in that case it's better to source `.bashrc` explicitly in the generated PBS file, instead of relying on exporting with `-V` IMO.

PR #161, opened 2017-03-15
Updates to make smart-dispatch Python 3 compatible

jfsantos:
I made a few updates to make smart-dispatch compatible with Python 3, namely replacing all print statement calls by the print function, converting a bytestring to a string and using `open` instead of `file` to detect if the argument of the option `-f` is a file. I did not do extensive testing yet but this does not break it with Python 2 or 3 when running a basic test (`smart_dispatch.py -qtest launch nvidia-smi`).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,6 @@ setup(
     license='LICENSE.txt',
     description='An easy to use job launcher for supercomputers with PBS compatible job manager.',
     long_description=open('README.md').read(),
-    install_requires=['psutil>=1'],
+    install_requires=['psutil>=1', 'future'],
     package_data={'smartdispatch': ['config/*.json']}
 )
--- a/smartdispatch/utils.py
+++ b/smartdispatch/utils.py
@@ -1,5 +1,11 @@
 from __future__ import print_function
 
+try:
+    _unicode = unicode
+    _utf8 = lambda x: _unicode(x, 'UTF-8')
+except NameError:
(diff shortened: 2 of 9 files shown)


jfsantos:
There are just a few issues on Travis, I'm going to check them out soon.

setup.py
```diff
@@ -14,6 +14,6 @@
     license='LICENSE.txt',
     description='An easy to use job launcher for supercomputers with PBS compatible job manager.',
     long_description=open('README.md').read(),
-    install_requires=['psutil>=1'],
+    install_requires=['psutil>=1', 'future'],
```
MarcCote: Didn't know about that library. Neat!

smartdispatch/utils.py
```diff
@@ -1,7 +1,21 @@
+from __future__ import print_function
+
+from builtins import str
+from builtins import input
+from builtins import map
+from builtins import range
+
+try:
+    _unicode = unicode
+    _utf8 = lambda x: _unicode(x, 'UTF-8')
```
MarcCote: I would have thought this sort of "hacks" isn't needed anymore thanks to the `future` library?

jfsantos: You're right. I did this before adding `future` as a dependency because I thought I could get around it without additional dependencies... until I got to the tests. I'll change it to use `future.builtins.str` instead.

PR #166, opened 2017-08-09
Test for pbs files on slurm

aalitaiga:


smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+from subprocess import Popen, PIPE
+from nose.tools import assert_equal, assert_true
+
+pbs_string = """\
+#!/usr/bin/env /bin/bash
+
+#PBS -N arrayJob
+#PBS -o arrayJob_%A_%a.out
+#PBS -t 1-2
```
bouthilx: We can ignore arrays as long as SmartDispatch is not using them.

smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -0,0 +1,133 @@
...
+    def test_constraint(self):
+        test_param(
+            ["gpu6gb", "gpu8gb"],
+            "PBS -l proc={}",
+            "feature",
+            pbs_string
+        )
+
+#
+# pbs_string2 = """\
```
bouthilx: Why a second PBS template?

smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -0,0 +1,133 @@
...
+def test_param(param_array, com, flag, string=pbs_string):
+    for param in param_array:
+        command = pbs_string.format(
+            string.format(com.format(param))
+        )
+        with open("test.pbs", "w") as text_file:
+            text_file.write(command)
+        process = Popen("sbatch test.pbs", stdout=PIPE, stderr=PIPE, shell=True)
+        stdout, stderr = process.communicate()
+        assert_true("Submitted batch job" in stdout)
```
bouthilx: You can define the function as a Method of TestSlurm and use ```self.assertIn``` rather than ```assert_true```

smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -0,0 +1,133 @@
...
+        job_id = stdout.split(" ")[-1].strip()
+
+        time.sleep(0.25)
+        process = Popen("squeue -u $USER -j {} -O {}".format(job_id, flag), stdout=PIPE, stderr=PIPE, shell=True)
+        stdout, stderr = process.communicate()
+        job_params = [c.strip() for c in stdout.split("\n")[1:] if c != '']
+        # import ipdb; ipdb.set_trace()
+        assert_true(all(p == param for p in job_params))
```
bouthilx: Again, this would give a more detailed error message ```self.assertSequenceEqual(job_params, (param for _ in job_params))```

smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -0,0 +1,133 @@
...
+
+class TestSlurm(unittest.TestCase):
+
+    def tearDown(self):
+        process = Popen("rm *.out test.pbs", stdout=PIPE, stderr=PIPE, shell=True)
+        stdout, stderr = process.communicate()
```
bouthilx: I think it'd be better to use glob instead of Popen. 

```python
for file_name in (glob.glob('*.out') + ["test.pbs"]):
    os.remove(file_name)
```

smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -0,0 +1,114 @@
...
+nvidia-smi
+"""
+
+sbatch_string = """\
+#!/usr/bin/env -i /bin/zsh
+
+#SBATCH --job-name=arrayJob
+#SBATCH --output=arrayJob_%A_%a.out
+#SBATCH --time=01:00:00
+#SBATCH --gres=gpu
```
bouthilx: -l naccelerators=x will give gpu:x with ```squeue -o gres```, so you should use --gres=gpu:x in sbatch file too. You should not test for 1 only, give it a random integer too.

smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -0,0 +1,114 @@
...
+#SBATCH --constraint=gpu6gb
```
bouthilx: gpu:x and constraint should be formatable in sbatch string since we want to try different values.

smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -48,26 +46,25 @@ def tearDown(self):
...
 
-    def _test_param(self, param_array, com, flag, string=pbs_string):
-        for param in param_array:
-            command = pbs_string.format(
-                string.format(com.format(param))
+    def _test_param(self, param_array, command, flag, string=pbs_string, output_array=None):
+        output_array = output_array or param_array
+        for param, output in zip(param_array, output_array):
+            com = pbs_string.format(
+                string.format(command.format(param))
```
bouthilx: Just for clarity:
command -> command_template
com -> param_command

smartdispatch/job_generator.py
```diff
@@ -171,3 +181,13 @@ def _add_cluster_specific_rules(self):
         for pbs in self.pbs_list:
             # Remove forbidden ppn option. Default is 2 cores per gpu.
             pbs.resources['nodes'] = re.sub(":ppn=[0-9]+", "", pbs.resources['nodes'])
+
+class SlurmClusterGenerator(JobGenerator):
+
+    def _add_cluster_specific_rules(self):
+        for pbs in self.pbs_list:
+            node_resource = pbs.resources.pop('nodes')
```
bouthilx: It doesn't seem wise to pop ['nodes'] if there is other information than gpus and ppn. I would rather substitute "gpus=[0-9]+:" with "". However we would need to check if resources['nodes'] becomes empty afterwards and remove it if so. It is very likely to happen, but it is necessary if we want to keep other informations when present.

bouthilx: It doesn't seem wise to pop ['nodes'] if there is other information than gpus and ppn. I would rather substitute "gpus=[0-9]+:" with "". However we would need to check if resources['nodes'] becomes empty afterwards and remove it if so. It is very likely to happen, but it is necessary if we want to keep other informations when present.

smartdispatch/pbs.py
```diff
@@ -62,6 +64,18 @@ def add_options(self, **options):
 
             self.options["-" + option_name] = option_value
 
+    def add_sbatch_options(self, **options):
+        """ Adds sbatch options to this PBS file.
+
+        Parameters
+        ----------
+        **options : dict
+            each key is the name of a SBATCH option (see `Options`)
```
bouthilx: There is no Options section in this docstring

smartdispatch/pbs.py
```diff
@@ -62,6 +64,18 @@ def add_options(self, **options):
...
+        """
+
+        for option_name, option_value in options.items():
+            self.sbatch_options[option_name] = option_value
```
bouthilx: We would need to add back the "-"/"--" here if we drop it in ```add_sbatch_flags```. 
```python
if len(option_name) == 1:
    "-" + option_name 
else:
    "--" + option_name
```

smartdispatch/job_generator.py
```diff
@@ -73,6 +73,16 @@ def add_pbs_flags(self, flags):
             pbs.add_resources(**resources)
             pbs.add_options(**options)
 
+    def add_sbatch_flags(self, flags):
+        options = {}
+
+        for flag in flags:
+            split = flag.find('=')
+            options[flag[:split]] = flag[split+1:]
```
bouthilx: It would be more consistent with the PBS options treatment if we drop the "-"/"--" in front of the option names. I would also raise a ValueError if neither are present.

smartdispatch/tests/pbs_slurm_test.py
```diff
@@ -39,6 +39,13 @@
 nvidia-smi
 """
 
+# Checking which cluster is running the tests first
+process = Popen("sacctmgr list cluster", stdout=PIPE, stderr=PIPE, shell=True)
```
bouthilx: would subprocess.check_output be simpler to use? It's used widely in this repo.

bouthilx: Also I would write functions for qstat and slurm cluster detections. Something like 
```python
def get_qstat_cluster_name():
    """
    ... Some doc ...

    Raises
    ------
    OSError
        The cluster may not have qstat installed.
    """
    output = Popen(["qstat", "-B"], stdout=PIPE).communicate()[0]
    return output.split('\n')[2].split(' ')[0]


def get_slurm_cluster_name():
    """
    ... Some doc ...
    Raises
    ------
    OSError
        The cluster may not have slurm installed.
    """
    # ... Some code ...
```

Those would be used in ```detect_cluster()```

smartdispatch/utils.py
```diff
@@ -115,6 +115,7 @@ def detect_cluster():
         output = Popen(["qstat", "-B"], stdout=PIPE).communicate()[0]
     except OSError:
         # If qstat is not available we assume that the cluster is unknown.
+        # TODO: handle MILA + CEDAR + GRAHAM
```
bouthilx: You can insert the function ```get_slurm_cluster_name()``` here for reference, but let it still return None. We don't want to add support for specific clusters now. That will be another ticket.

bouthilx:
@aalitaiga You should rename pbs_slurm_test.py otherwise Travis will keep reporting failures. We could add --ignore-files=<name> in .travis.yml but we don't want those tests to run locally neither. They are only usefull if we want to tests new settings on the slurm configuration, so they should only be ran alone. A name like verify_slurms_pbs_wrapper.py would be self-explanatory.

smartdispatch/job_generator.py
aalitaiga: If go from "nodes=1:ppn=2:gpus=2" to "nodes=1" is it okay or I should remove nodes completely in this case?

bouthilx: It is ok

smartdispatch/job_generator.py
```diff
@@ -78,7 +78,12 @@ def add_sbatch_flags(self, flags):
 
         for flag in flags:
             split = flag.find('=')
-            options[flag[:split]] = flag[split+1:]
+            if flag.startswith('--'):
+                options[flag[2:split]] = flag[split+1:]
```
bouthilx: This would be simpler.

```python
if not flag.startswith("-"):
  raise ValueError("Invalid SBATCH flag ({})".format(flag))

options[flag[:split].lstrip("-")] = flag[split+1:]
```

smartdispatch/job_generator.py
```diff
@@ -182,12 +187,13 @@ def _add_cluster_specific_rules(self):
...
 
     def _add_cluster_specific_rules(self):
         for pbs in self.pbs_list:
-            node_resource = pbs.resources.pop('nodes')
-            gpus = re.match(".*gpus=([0-9]+)", node_resource).group(1)
-            ppn = re.match(".*ppn=([0-9]+)", node_resource).group(1)
+            gpus = re.match(".*gpus=([0-9]+)", pbs.resources['nodes']).group(1)
+            ppn = re.match(".*ppn=([0-9]+)", pbs.resources['nodes']).group(1)
+            pbs.resources['nodes'] = re.sub("ppn=[0-9]+", "", pbs.resources['nodes'])
+            pbs.resources['nodes'] = re.sub(":gpus=[0-9]+", "", pbs.resources['nodes'])
```
bouthilx: You don't look for empty pbs.resources['nodes'] because there is always nodes=1 remaining? That would be fine in my opinion, I just want to be sure to understand for which reason you do not check.

smartdispatch/pbs.py
bouthilx: Isn't there a duplicate at line 81? You can change it to option_name = "-" + option_name and leave line 81 like it is.

smartdispatch/utils.py
```diff
@@ -115,7 +115,7 @@ def detect_cluster():
         output = Popen(["qstat", "-B"], stdout=PIPE).communicate()[0]
     except OSError:
         # If qstat is not available we assume that the cluster is unknown.
-        # TODO: handle MILA + CEDAR + GRAHAM
+        cluster_name = get_slurm_cluster_name()
```
bouthilx: Please leave the TODO there. Otherwise, it is not obvious why it still returns None.

smartdispatch/utils.py
```diff
@@ -131,6 +131,11 @@ def detect_cluster():
         cluster_name = "hades"
     return cluster_name
 
+def get_slurm_cluster_name():
```
bouthilx, line 162: We need a unit test for this

smartdispatch/job_generator.py
```diff
@@ -73,6 +73,21 @@ def add_pbs_flags(self, flags):
             pbs.add_resources(**resources)
             pbs.add_options(**options)
 
+    def add_sbatch_flags(self, flags):
```
bouthilx, line 78: We need a unit test for add_sbatch_flags

smartdispatch/job_generator.py
```diff
@@ -171,3 +186,14 @@ def _add_cluster_specific_rules(self):
         for pbs in self.pbs_list:
             # Remove forbidden ppn option. Default is 2 cores per gpu.
             pbs.resources['nodes'] = re.sub(":ppn=[0-9]+", "", pbs.resources['nodes'])
+
+class SlurmJobGenerator(JobGenerator):
+
+    def _add_cluster_specific_rules(self):
```
bouthilx: We need a unit test for this

scripts/smart-dispatch
```diff
@@ -206,6 +209,7 @@ def parse_arguments():
 
     parser.add_argument('-p', '--pool', type=int, help="Number of workers that will be consuming commands. Default: Nb commands")
     parser.add_argument('--pbsFlags', type=str, help='ADVANCED USAGE: Allow to pass a space seperated list of PBS flags. Ex:--pbsFlags="-lfeature=k80 -t0-4"')
+    parser.add_argument('--sbatchFlags', type=str, help='ADVANCED USAGE: Allow to pass a space seperated list of SBATCH flags. Ex:--sbatchFlags="-qos=high --output=file.out"')
```
bouthilx: We need a unit test for this.

bouthilx:
We need unit-tests for 
- [x] add_sbatch_flags
- [x] add_sbatch_options
- [x] ~~command line option~~
- [x] job generator pbs file replacements (pnn, gpus, etc)

Make sure to test that what should pass, pass and what should fail fails. We need to try many different things to test the command line options. For instance, we need to try adding bad options or adding options already included by job generator. We need to make sure to test every cracks as this seems to me as an error-prone setup (replacing options for a framework by options a wrapper of another framework will recognize as what we want it to be for the latter framework :sweat:).

smartdispatch/job_generator.py
aalitaiga: Yes it is the reason

smartdispatch/utils.py
aalitaiga: There's no unittest for 'detect_cluster' should I make one for 'get_cluster_name'?

bouthilx: There should be one for each functions.

smartdispatch/job_generator.py
```diff
@@ -79,11 +79,13 @@ def add_sbatch_flags(self, flags):
         for flag in flags:
             split = flag.find('=')
             if flag.startswith('--'):
-                options[flag[2:split]] = flag[split+1:]
-            elif flag.startswith('-'):
-                options[flag[1:split]] = flag[split+1:]
+                if split == -1:
+                    raise ValueError("Invalid SBATCH flag ({})".format(flag))
```
bouthilx: We know it's invalid probably because it should be a single - or the = is missing. The error message should give a hint about it.

smartdispatch/job_generator.py
```diff
@@ -79,11 +79,13 @@ def add_sbatch_flags(self, flags):
...
+                options[flag[:split].lstrip("-")] = flag[split+1:]
+            elif flag.startswith('-') and split == -1:
+                options[flag[1:2]] = flag[2:]
```
bouthilx: simply flag[1]? :stuck_out_tongue:

smartdispatch/tests/test_job_generator.py
```diff
@@ -129,6 +130,32 @@ def test_add_pbs_flags_invalid(self):
...
+    def test_add_sbatch_flags(self):
+        for flag in self.sbatch_flags:
+            yield self._test_add_sbatch_flags, [flag]
+
+        yield self._test_add_sbatch_flags, [flag]
+
+    def test_add_sbatch_flag_invalid(self):
+        invalid_flags = ["--qos high", "gpu", "-lfeature=k80"]
+        for flag in invalid_flags:
+            assert_raises(ValueError, self._test_add_sbatch_flags, "--qos high")
```
bouthilx: you pass the same flag each time.

smartdispatch/tests/test_job_generator.py
```diff
@@ -244,6 +271,28 @@ def test_pbs_split_2_job_nb_commands(self):
...
+        self.nb_cores_per_node = 1
+        self.nb_gpus_per_node = 2
+        self.queue = Queue("slurm", "mila", self.walltime, self.nb_cores_per_node, self.nb_gpus_per_node, self.mem_per_node)
+
+        self.commands = ["echo 1", "echo 2", "echo 3", "echo 4"]
+        job_generator = SlurmJobGenerator(self.queue, self.commands)
+        self.pbs = job_generator.pbs_list
+
+    def test_ppn_ncpus(self):
+        assert_true("ppn" not in str(self.pbs[0]))
```
bouthilx: Don't forget to make a dummy mock of self._add_cluster_specific_rules() to test that otherwise ppn was indeed present in the PBS and ncpus was not.

README.md
```diff
@@ -74,3 +74,11 @@ python my_script.py 9
 Jobs that did not terminate properly, for example, it exceeded the walltime, can be resumed using the {batch_id} given to you upon launch. Of course, all this assuming your script is resumable.
 
 *Note: Jobs are always in a batch, even if it's a batch of one.*
+
+### SLURM clusters
+
+Smartdispatch can also run on SLURM clusters
+All features like `--gpusPerNode` or `--coresPerNode` are supported
+However you need to pass SLURM specific features like `--qos` or `--output`as sbatch flach with:
+`smart-dispatch -q qtest@mp2 launch python my_script.py [1:4] --sbatchFlags="--qos=high -ofile.out`
```
bouthilx: -o Is defined by SmartDispatch, it's not a good idea to redefine it here. At least, we should not give it as a basic example.

README.md
```diff
@@ -74,3 +74,11 @@ python my_script.py 9
...
+If you use the full name of a feature you have to separate the feature and its value by a = like `--qos=high` if not you have to append the features and its value ex: `-ofile.out` is valid, not `-o file.out`
```
bouthilx: I feel it is a bit verbose but not so clear. What about this:
> For simplicity, --sbatchFlags supports short and long option definitions only with the following syntax:
> -Cgpu6gb or --constraint=gpu6gb
> For comparison, this would be invalid:
> -C gpu6gb or --constraint gpu6gb.

README.md
```diff
@@ -74,3 +74,11 @@ python my_script.py 9
 Jobs that did not terminate properly, for example, it exceeded the walltime, can be resumed using the {batch_id} given to you upon launch. Of course, all this assuming your script is resumable.
 
 *Note: Jobs are always in a batch, even if it's a batch of one.*
+
+### SLURM clusters
+
+Smartdispatch can also run on SLURM clusters
```
bouthilx: dot

README.md
```diff
@@ -74,3 +74,11 @@ python my_script.py 9
...
+All features like `--gpusPerNode` or `--coresPerNode` are supported
```
bouthilx: dot again. Look at the REAME.md file on github, you will see how it is rendered.

--gpusPerCommand and --coresPerCommand would make more sense for most users than --gpusPerNode. Also, they are quite similar. I would give different ones, like --gpusPerComand and --mem for instance. Wait, do we have support for mem? Argh! let me take a look at this. :disappointed:

Issue #96, opened 2015-06-17
When msub failed to launch a job we should have some feedback


mgermain:
For example, let say we launch a job on colosse and smartdispatch does not add the #PBS -A ycy-xxx-xx then msub will warn that the config is not proper but smartdispatch totally swallow the warning and we don't see it.

Issue #155, opened 2017-02-02
Handle resources error gracefully.


mgermain:
If you ask for more resources than what is available on the requires queue we just crash. We should handle that gracefully and throw a proper error message.

See #124 for example.

PR #167, opened 2017-09-13
Handle resources error gracefully. (#155)

Dutil:
Adresses #155 .


smartdispatch/tests/test_smartdispatch_script.py
```diff
@@ -0,0 +1,49 @@
...
+        shutil.rmtree(self._base_dir)
+
+    def test_gpu_check(self):
+
+        argv = ['-x', '-g', '2', '-G', '1', '-C', '1', '-q', 'random', '-t', '10' ,'launch', 'echo', 'testing123']
+
+        with self.assertRaises(SystemExit) as context:
+            smartdispatch_script.main(argv=argv)
+
+        self.assertTrue(context.exception.code, 2)
```
bouthilx: We also want to check that it can pass.

smartdispatch/tests/test_smartdispatch_script.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def test_cpu_check(self):
+
+        argv = ['-x', '-c', '2', '-C', '1', '-G', '1', '-t', '10', '-q', 'random', 'launch', 'echo', 'testing123']
+
+        with self.assertRaises(SystemExit) as context:
+            smartdispatch_script.main(argv=argv)
+
+        self.assertTrue(context.exception.code, 2)
```
bouthilx: idem

smartdispatch/tests/test_smartdispatch_script.py
```diff
@@ -0,0 +1,49 @@
...
+    def test_launch_job_check(self, mock_check_output):
+
+        mock_check_output.side_effect = subprocess.CalledProcessError(1, 1, "A wild error appeared!")
+        argv = ['-t', '0:0:1', '-G', '1', '-C', '1', '-q', 'random', 'launch', 'echo', 'testing123']
+
+        try:
+            with self.assertRaises(SystemExit) as context:
+                smartdispatch_script.main(argv=argv)
+
+                self.assertTrue(context.exception.code, 2)
```
bouthilx: idem

smartdispatch/tests/test_smartdispatch_script.py
```diff
@@ -0,0 +1,49 @@
...
+
+        except subprocess.CalledProcessError:
+            self.fail("smartdispatch_script.main() raised CalledProcessError unexpectedly!")
```
bouthilx: Do you know if self.fail will keep the stack trace from CalledProcessError? I know it is mocked so there won't be interesting information in the message itself, but the stacktrace would point to the unexpected call to check_output.

tests/test_smart_dispatch.py
```diff
@@ -95,6 +101,31 @@ def test_main_launch_with_cores_command(self):
         assert_equal(exit_status_100, 2)        
         assert_true(os.path.isdir(self.logs_dir))
 
+    def test_main_launch_with_gpus_command(self):
+        # Actual test
+        exit_status_100 = call(self.launch_command_with_gpus.format(gpus=100), shell=True)
+
+        # Test validation
```
bouthilx: Why not testing 0 gpus too? Because it passes? Then we could assert_equal(test exit_status_0, 0).

tests/test_smart_dispatch.py
```diff
@@ -95,6 +101,31 @@ def test_main_launch_with_cores_command(self):
...
+
+        #For this test, we won't call the script directly, since we want to mock subprocess.check_output
+        mock_check_output.side_effect = subprocess.CalledProcessError(1, 1, "A wild error appeared!")
+        argv = ['-t', '0:0:1', '-G', '1', '-C', '1', '-q', 'random', 'launch', 'echo', 'testing123']
+
+        try:
+            with self.assertRaises(SystemExit) as context:
+                smartdispatch_script.main(argv=argv)
+
+                self.assertTrue(context.exception.code, 2)
```
bouthilx: idem for passing tests

tests/test_smart_dispatch.py
```diff
@@ -95,6 +101,31 @@ def test_main_launch_with_cores_command(self):
...
+
+        except subprocess.CalledProcessError:
+            self.fail("smartdispatch_script.main() raised CalledProcessError unexpectedly!")
```
bouthilx: idem for stacktrace

smartdispatch/smartdispatch_script.py
```diff
@@ -0,0 +1,249 @@
...
+            sys.stderr.write("smart-dispatch: error: The launcher wasn't able the launch the job(s) properly. Maybe the pbs file(s) generated were invalid. The following error message was returned: \n{}".format(e.output))
+            sys.exit(2)
+
+    print "\nLogs, command, and jobs id related to this batch will be in:\n {smartdispatch_folder}".format(smartdispatch_folder=path_job)
+
+
+def parse_arguments(argv=None):
+
+    if argv is None:
+        argv = sys.argv[1:]
```
bouthilx: That's useless, parser.parse_args(None) will do this internally.

bouthilx:
Note for myself, this PR should also fix #96.

smartdispatch/tests/test_smartdispatch_script.py
Dutil: No it doesn't. Gonna add the stack trace to the assert message.

bouthilx: I think it would be better to reraise the error with the message changed than adding the stacktrace as a string to the message (if I understand correctly what you intend).

I would use [six.reraise](https://pythonhosted.org/six/#six.reraise). You can look [here](https://github.com/mila-udem/blocks/blob/master/blocks/utils/utils.py#L96) for a usage example.

smartdispatch/tests/test_smartdispatch_script.py
bouthilx: I'm getting confused here. You expect the function to raise SystemExit, but you mocked adding side_effect CalledProcessError with exception code 1. Why would it rather raise SystemExit?

By the way, what's the difference with test_smart_dispatch.TestSmartdispatcher.test_launch_job_check?

Dutil: So it raise a CalledProcessError, which is catch here: https://github.com/Dutil/smartdispatch/blob/master/smartdispatch/smartdispatch_script.py#L193
and then quit. I don't reraise the exception in the script, since qsub/msub already print the error message and the stack trace is not really helpful.

And yeah, just realized that I had the same test 2 times, and forgot to remove one. In fact I think I should remove the smartdispatch/tests/test_smartdispatch_script.py entirely, since the tests that I've add in test_smart_dispatch test the same thing, just in different ways.

bouthilx: >So it raise a CalledProcessError, which is catch here: https://github.com/Dutil/smartdispatch/blob/master/smartdispatch/smartdispatch_script.py#L193
and then quit. I don't reraise the exception in the script, since qsub/msub already print the error message and the stack trace is not really helpful.

Oh right! Thanks for the clarification

> And yeah, just realized that I had the same test 2 times, and forgot to remove one. In fact I think I should remove the smartdispatch/tests/test_smartdispatch_script.py entirely, since the tests that I've add in test_smart_dispatch test the same thing, just in different ways.

I agree, those tests are somehow duplicates. Anyhow, I would add tests in test_smart_dispatch which call smartdispatch.main directly. That will make it easier to test the error messages.

Dutil: Ok, should I just move test_gpu_check and test_cpu_check to test_smart_dispatch.TestSmartdispatcher?

bouthilx: I would say so, yes.

bouthilx:
@Dutil You commit's title says "Removing duplicate test" but I don't see anything removed. Did I miss something?

smartdispatch/tests/test_smartdispatch_script.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self.assertTrue(context.exception.code, 2)
+
+        # Test if the test pass
+        argv[2] = '0'
+
+        try:
+            smartdispatch_script.main(argv=argv)
+        except SystemExit as e:
+            self.fail("The command failed the check, but it was supposed to pass.")
```
bouthilx: What happens if you just let it crash on smartdispatch_script.main()? Is it bad because of SystemExit? Do you lose the stacktrace or a valuable error message by doing self.fail?

Dutil:
@bouthilx Yeah sorry, I've remove the other file now.

tests/test_smart_dispatch.py
```diff
@@ -95,6 +103,87 @@ def test_main_launch_with_cores_command(self):
...
+
+        self.assertTrue(context.exception.code, 2)
+
+        # Test if the test pass
+        argv[2] = '1'
+
+        try:
+            smartdispatch_script.main(argv=argv)
+        except SystemExit as e:
+            self.fail("The command failed the check, but it was supposed to pass.")
```
bouthilx: What happens if you just let it crash on smartdispatch_script.main()? Is it bad because of SystemExit? Do you lose the stacktrace or a valuable error message by doing self.fail?

Dutil: Yes, we lose the the information of the stacktrace when using the self.fail. Otherwise the test crash and we get an error (the rest of the tests runs normally).

I used the self.fail instead of letting the script crash, because I felt it was more inline with what the test was supposed to do (i.e. the test failed because a SystemExit  was thrown, it's not really an unexpected error). But it's right that we lose information that way. Should I simply let it crash instead?

bouthilx: I think it is important when a unit test fails that the error message is informative enough that we can easily pin-point the problem and fix it. In this case, I believe both SystemExit and self.fail are just as bad, because SystemExit doesn't write down the stack trace nor informative message about the source of the problem, right? If this is true, then I would do like the other tests where the error is reraised as another exception type. In this case, the reraise procedures should be refactored as a single function to avoid code duplication.

smartdispatch/utils.py
```diff
@@ -136,3 +136,26 @@ def get_launcher(cluster_name):
         return "msub"
     else:
         return "qsub"
+
+def get_advice(cluster_name):
+
+    helios_advice = """On Helios, don't forget that the queue gpu_1, gpu_2, gpu_4 and gpu_8 give access to a specific amount of gpus.
```
bouthilx: pep8 please :cry: :stuck_out_tongue_winking_eye:

smartdispatch/utils.py
```diff
@@ -136,3 +136,26 @@ def get_launcher(cluster_name):
...
+For more advices, please refer to the official documentation: 'https://wiki.calculquebec.ca/w/Helios/en'"""    
+    mammouth_advice = "On Mammouth, please refer to the official documentation for more information: 'https://wiki.ccs.usherbrooke.ca/Accueil/en'"
+    hades_advice = """On Hades, don't forget that the queue name '@hades' needs to be use.
```
bouthilx: needs to be *used

smartdispatch/smartdispatch_script.py
```diff
@@ -0,0 +1,249 @@
...
+    print "Batch UID:\n{batch_uid}".format(batch_uid=jobname)
+    if not args.doNotLaunch:
+        
+        try:
+            launch_jobs(LAUNCHER if args.launcher is None else args.launcher, pbs_filenames, CLUSTER_NAME, path_job)
+        except subprocess.CalledProcessError as e:
+
+            cluster_advice = utils.get_advice(CLUSTER_NAME)
+
+            sys.stderr.write("smart-dispatch: error: The launcher wasn't able the launch the job(s) properly. The following error message was returned: \n\n{}\n\nMaybe the pbs file(s) generated were invalid. {}\n\n".format(e.output, cluster_advice))
```
bouthilx: I won't ask you to _pep8ify_ all code, but please do so for the lines you editied like this one.

smartdispatch/smartdispatch_script.py
```diff
@@ -0,0 +1,249 @@
...
+
+    # Check that requested core number does not exceed node total
+    if args.coresPerCommand > queue.nb_cores_per_node:
+        sys.stderr.write("smart-dispatch: error: coresPerCommand exceeds nodes total: asked {req_cores} cores, nodes have {node_cores}\n"
+                         .format(req_cores=args.coresPerCommand, node_cores=queue.nb_cores_per_node))
+        sys.exit(2)
+
+    # Check that requested gpu number does not exceed node total
+    if args.gpusPerCommand > queue.nb_gpus_per_node:
+        sys.stderr.write("smart-dispatch: error: gpusPerCommand exceeds nodes total: asked {req_gpus} gpus, nodes have {node_gpus}\n"
```
bouthilx: Maybe a little advice would be helpful here too. Some users might be confused to get a message saying there is only 1 GPU per node available if they know nodes have more than that. They might not know that specific queues can limit the number of GPUs available (on helios for instance).

smartdispatch/utils.py
```diff
@@ -136,3 +136,26 @@ def get_launcher(cluster_name):
         return "msub"
     else:
         return "qsub"
+
+def get_advice(cluster_name):
```
bouthilx: I think it would be preferable to have the advices outside of get_advice, following the constant variable naming convention (HELIOS_ADVICE, MAMMOUTH_ADVICE, etc).

tests/test_smart_dispatch.py
```diff
@@ -3,11 +3,35 @@
...
 from subprocess import call
-
+import subprocess
 from nose.tools import assert_true, assert_equal
+from smartdispatch import smartdispatch_script
+import six
+import sys
+import traceback
+
+def rethrow_exception(exception, new_message):
```
bouthilx: What about moving this helper function to utils.py? utils/testing.py would be even better actually but I would leave utils refactoring for later.

tests/test_smart_dispatch.py
```diff
@@ -23,17 +47,23 @@ def setUp(self):
         self.nb_commands = len(self.commands)
 
         scripts_path = abspath(pjoin(os.path.dirname(__file__), os.pardir, "scripts"))
-        self.smart_dispatch_command = '{} -C 1 -q test -t 5:00 -x'.format(pjoin(scripts_path, 'smart-dispatch'))
+        self.smart_dispatch_command = '{} -C 1 -G 1 -q test -t 5:00 -x'.format(pjoin(scripts_path, 'smart-dispatch'))
```
bouthilx, line 28: Do we still have tests without GPUs? We need to test both settings; With and without GPUs.

Dutil: So I just remembered why I added that. the default value for 'gpusPerCommand' is 1, but for 
 'gpusPerNode' the default value is 0, when the queue is undefined (like the queue 'test', that is used for the tests). And that triggers the gpu check.

So in order to have those tests pass, we need to either have the -G 1 options, change the check to let it pass if gpusPerNode == 0, or change the gpusPerCommand default to 0.

smartdispatch/smartdispatch_script.py
Dutil: So to show the same advice as when qsub/msub fail?

tests/test_smart_dispatch.py
bouthilx: Then we should have tests for -g 0 -G 0 and -g 0. Do we?

smartdispatch/smartdispatch_script.py
bouthilx: Yeah that's not so nice. Maybe just had something like: Make sure you specified the correct queue.

tests/test_smart_dispatch.py
```diff
@@ -150,7 +128,22 @@ def test_gpu_check(self):
         argv[2] = '1'
         smartdispatch_script.main(argv=argv)
 
-    @rethrow_exception(SystemExit, "smartdispatch_script.main() raised SystemExit unexpectedly.")
+        # Test if we don't have gpus. (and spicified in script).
```
bouthilx: spicified :stuck_out_tongue_closed_eyes:

smartdispatch/utils.py
```diff
@@ -136,3 +171,25 @@ def get_launcher(cluster_name):
         return "msub"
     else:
         return "qsub"
+
+
+def rethrow_exception(exception, new_message):
+
+    def func_wraper(func):
+
+        def test_func(*args, **kwargs):
```
bouthilx: Wouldn't be better to use @functools.wraps(func) here for the stacktrace? I mean

```python
@functools.wraps(func)
def test_func(*args, **kwargs):
```

I just had a problem with stack trace because of decorators while implementing tests for Slurm clusters. I wonder if the same thing is happening here.

tests/test_smart_dispatch.py
bouthilx: Aren't the -g 0 -G 0 and -g 0 tests still missing?

tests/test_smart_dispatch.py
```diff
@@ -95,6 +103,84 @@ def test_main_launch_with_cores_command(self):
...
+        with self.assertRaises(SystemExit) as context:
+            smartdispatch_script.main(argv=argv)
+
+        self.assertTrue(context.exception.code, 2)
+
+        # Test if the test pass
+        argv[2] = '1'
+        smartdispatch_script.main(argv=argv)
+
+        # Test if we don't have gpus. (and spicified in script).
```
bouthilx: Looks like you should do ```grep "spicified"``` :stuck_out_tongue_closed_eyes:

tests/test_smart_dispatch.py
```diff
@@ -95,6 +103,84 @@ def test_main_launch_with_cores_command(self):
...
+        argv[2] = '0'
+        argv[4] = '0'
+        smartdispatch_script.main(argv=argv)
+
+        # Don't have gpus, but the user specofy 1 anyway.
```
bouthilx: A variant! Specofy! :upside_down_face:

tests/test_smart_dispatch.py
Dutil: In my defence, it's the same line/mistake as the other one :stuck_out_tongue_winking_eye:

smartdispatch/utils.py
Dutil: Just tried, it didn't change the stack trace.

tests/test_smart_dispatch.py
bouthilx: Coherence is important indeed

smartdispatch/utils.py
bouthilx: OK, well you can leave it there anyway.

tests/test_smart_dispatch.py
bouthilx: Wait, it's the same file? hahaha! :blush:

bouthilx:
@mgermain Ready for merge. I believe the decrease in coverage is due to the move of scripts/* to smartdispatch/*_script.py. The unit-tests for the proposed changes seems to have a proper coverage in my opinion.

